PR #4, opened 2020-05-03
Allow commands to be executed without case sensitivity.

Jason2605:
# Execute commands regardless of case
## Summary
This PR changes the command execution in the bot to ignore case when executing a command.

### DIFF
--- a/test/factories/CommandFactoryTest.ts
+++ b/test/factories/CommandFactoryTest.ts
@@ -1,6 +1,5 @@
 import { expect } from "chai";
 
-// @ts-ignore - TS does not like MockDiscord not living in src/
 import CommandFactory from "../../src/factories/CommandFactory";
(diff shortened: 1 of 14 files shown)


LamboCreeper:
Looks good, although I'd really appreciate it if you wrote tests.

test/factories/CommandFactoryTest.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from "chai";
+
+// @ts-ignore - TS does not like MockDiscord not living in src/
```
LamboCreeper: Is this comment meant to be here?

LamboCreeper (approved):

test/factories/CommandFactoryTest.ts
Jason2605: Nope, good spot

PR #6, opened 2020-05-03
Event Handling

LamboCreeper:
#### Overview
- Added an `EventHandler` abstract to base handlers off of
- Added `CommandParserHandler` to handle commands
- Moved logic for getting files in a directory into a util instead of living in the `CommandFactory`
- Added logic to `app.ts` for loading the helpers on start

### DIFF
--- a/src/event/handlers/CommandParserHandler.ts
+++ b/src/event/handlers/CommandParserHandler.ts
@@ -1,0 +1,27 @@
+import { Constants, Message } from "discord.js";
+import { COMMAND_PREFIX } from "../../config.json";
+import EventHandler from "../../abstracts/EventHandler";
+import CommandFactory from "../../factories/CommandFactory";
+
+class CommandParserHandler extends EventHandler {
+	constructor() {
+		super(Constants.Events.MESSAGE_CREATE);
+	}
+
+	async handle(message: Message): Promise<void> {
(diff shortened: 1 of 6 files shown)


src/event/handlers/CommandParserHandler.ts
```diff
@@ -0,0 +1,27 @@
...
+
+class CommandParserHandler extends EventHandler {
+	constructor() {
+		super(Constants.Events.MESSAGE_CREATE);
+	}
+
+	async handle(message: Message): Promise<void> {
+		const commandFactory = new CommandFactory();
+
+		await commandFactory.loadCommands();
```
AtrumAmantus: Create & Load the factory in the constructor, saving it to a class property, so that the constructor is only created once, when the `CommandParserHandler` is created.  Then on subsequent message events, it is only utilized, and not created each time.

LamboCreeper: Hmm, just tried to do this, however, because `CommandFactory.loadCommands()` is asynchronous it won't let me put it in the `constructor`. Any ideas?

LamboCreeper: Nevermind, @Jason2605 and I worked it out.

Jason2605 (approved):

AtrumAmantus (approved):

PR #24, opened 2020-05-31
Added alias functionality

### LINKED ISSUE #17: "codeblocks" doesn't trigger the codeblock command
Sometimes you want to trigger the codeblock command, so you enter `codeblocks`. Nothing happens. Then you try `codeblock` and it works. It would be nice if both `codeblock` and `codeblocks` triggered this command.

ajsaraujo:
- Added `aliases` field to `CommandOptions` and `getAliases()` method to `Command`
- Added the aliases "codeblock" and "cb" to the codeblocks command
- Load aliases to the commands object in CommandFactory
- Created tests for aliases behaviour on `CommandFactoryTest.ts`

### DIFF
--- a/src/abstracts/Command.ts
+++ b/src/abstracts/Command.ts
@@ -5,13 +5,11 @@ abstract class Command {
 	private readonly name: string;
 	private readonly description: string;
 	private readonly options: CommandOptions | undefined;
-	private readonly aliases: Array<string>;
 
-	protected constructor(name: string, description: string, options?: CommandOptions, aliases: Array<string> = []) {
+	protected constructor(name: string, description: string, options?: CommandOptions) {
 		this.name = name;
 		this.description = description;
 		this.options = options;
-		this.aliases = aliases;
 	}
 
 	abstract async run(message: Message, args?: string[]): Promise<void>;
@@ -24,8 +22,8 @@ abstract class Command {
 		return this.description;
 	}
 
-	getAliases(): Array<string> {
-		return this.aliases;
+	getAliases(): string[] {
+		return this.options?.aliases || [];
 	}
 
 	isSelfDestructing(): boolean {
--- a/src/abstracts/Command.ts
+++ b/src/abstracts/Command.ts
@@ -5,11 +5,13 @@ abstract class Command {
 	private readonly name: string;
 	private readonly description: string;
 	private readonly options: CommandOptions | undefined;
+	private readonly aliases: Array<string>;
 
-	protected constructor(name: string, description: string, options?: CommandOptions) {
+	protected constructor(name: string, description: string, options?: CommandOptions, aliases: Array<string> = []) {
 		this.name = name;
 		this.description = description;
 		this.options = options;
+		this.aliases = aliases;
 	}
 
 	abstract async run(message: Message, args?: string[]): Promise<void>;
@@ -20,6 +22,10 @@ abstract class Command {
 
 	getDescription(): string {
 		return this.description;
+	}
+
+	getAliases(): Array<string> {
+		return this.aliases;
 	}
 
 	isSelfDestructing(): boolean {
--- a/test/MockCommandWithAlias.ts
+++ b/test/MockCommandWithAlias.ts
@@ -5,10 +5,11 @@ export default class MockCommandWithAlias extends Command {
 	constructor() {
 		super(
 			"mock-alias",
-			"Mock Command With Alias",
+			"Mock Command with Aliases",
 			{
-				aliases: ["mocky", "mockster"]
-			});
+				aliases: ["mocky", "mocko"]
+			}
+		);
 	}
 
 	async run(message: Message, args: string[]): Promise<void> {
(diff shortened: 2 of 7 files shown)


src/abstracts/Command.ts
```diff
@@ -5,11 +5,13 @@ abstract class Command {
 	private readonly name: string;
 	private readonly description: string;
 	private readonly options: CommandOptions | undefined;
+	private readonly aliases: Array<string>;
```
LamboCreeper: Can you use `string[]` over `Array<string>` please, it's just shorter and neater in my opinion?

src/abstracts/Command.ts
```diff
@@ -5,11 +5,13 @@ abstract class Command {
...
 
-	protected constructor(name: string, description: string, options?: CommandOptions) {
+	protected constructor(name: string, description: string, options?: CommandOptions, aliases: Array<string> = []) {
```
LamboCreeper: I think it might be useful to move aliases into the `CommandOptions`. Keeps the `constructor` shorter.

src/abstracts/Command.ts
```diff
@@ -22,6 +24,10 @@ abstract class Command {
 		return this.description;
 	}
 
+	getAliases(): Array<string> {
```
LamboCreeper: Again can this be `string[]` please? :)

test/commands/CodeblockCommandTest.ts
```diff
@@ -20,6 +20,13 @@ describe("CodeblockCommand", () => {
 
 			expect(command.getDescription()).to.equal("Shows a tutorial on how to use Discord's codeblocks.");
 		});
+
+		it("creates a command with correct aliases", () => {
+			const command = new CodeblockCommand();
+
+			expect(command.getAliases().includes("codeblocks")).to.equal(true);
+			expect(command.getAliases().includes("cb")).to.equal(true);
+		});
```
LamboCreeper: We should really mock a command for testing rather than using an instance of `CodeblockCommand`. What happens if the logic changes for this command or it gets removed?

LamboCreeper (changes requested):
Looks good, just a few changes, please. 

Thanks for adding the functionality! :)

src/abstracts/Command.ts
ajsaraujo: Nice, I didn't know about that syntax, it is cleaner indeed. I'm going to fix it :+1:

src/abstracts/Command.ts
ajsaraujo: Nice, I didn't notice you could do that either.

test/factories/CommandFactoryTest.ts
```diff
@@ -6,67 +6,67 @@ import MockCommand from "../MockCommand";
 import MockCommandWithAlias from "../MockCommandWithAlias";
 
 describe("CommandFactory", () => {
-    let factory: CommandFactory;
-    let commandName: string;
```
ajsaraujo, line 9: The linter was complaining about some indentation problems (Expected 2 tabs, found 8 spaces). I didn't want to fix it because it would mess with the entire file and it was rendering normally in my editor. However, when I pushed the changes, the indenting on Github was awful (as you can see in the deleted lines). Hence, I autofixed all indentation problems. Let me know if this was a bad call.

test/MockCommand.ts
```diff
@@ -3,7 +3,7 @@ import { Message } from "discord.js";
 
 export default class MockCommand extends Command {
 	constructor() {
-		super('mock', 'Mock Command');
+		super("mock", "Mock Command");
```
ajsaraujo, line 6: The linter was complaining about the single quotes.

test/factories/CommandFactoryTest.ts
```diff
@@ -3,47 +3,70 @@ import { expect } from "chai";
 import CommandFactory from "../../src/factories/CommandFactory";
 // @ts-ignore - TS does not like MockCommand not living in src/
 import MockCommand from "../MockCommand";
+import MockCommandWithAlias from "../MockCommandWithAlias";
```
ajsaraujo, line 7: This is weird. The linter was complaining about indentation problems "Expected 2 tabs, found 8 spaces". Autofixing would change the entire file, and it was rendering ok on my editor, so I left it like it was. However, when I pushed it to GitHub, the rendering was pretty weird, so I pushed another commit with the autofix. Let me know if this was a bad call.

test/MockCommandWithAlias.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default class MockCommandWithAlias extends Command {
+	constructor() {
+		super(
+			"mock-alias",
+			"Mock Command With Alias",
+			{
+				aliases: ["mocky", "mockster"]
+			});
+	}
```
LamboCreeper: Indentation seems a bit messed up here.
```suggestion
	constructor() {
		super(
			"mock-alias",
			"Mock Command With Alias",
			{
				aliases: ["mocky", "mockster"]
			}
		);
	}
```

test/factories/CommandFactoryTest.ts
LamboCreeper: Yeah, so your editor will show four spaces and a tab being the same width, but GitHub will show them differently.

LamboCreeper (changes requested):

LamboCreeper (approved):

PR #34, opened 2020-06-06
Added a command to search NPM packages.

### LINKED ISSUE #29: NPM Command
Get a link to an NPM package.

Command Usage: `?npm <package>`

If a package is not supplied, return an error stating the correct usage (similar to the `?rule` command). If a package is supplied, perform a GET request: if 200 is returned send the link as the package is valid, if not state that a package of the supplied name could not be found.

This command should function identically to the one in the old Discord bot (see [here](https://github.com/codesupport/old-discord-bot/blob/develop/commands/npm.js)).

ajsaraujo:


src/commands/NPMCommand.ts
```diff
@@ -0,0 +1,41 @@
...
+	constructor() {
+		super(
+			"npm",
+			"Displays a link to a given NPM package.",
+		);
+	}
+
+	async run(message: Message, args: string[]) {
+		const embed = new MessageEmbed();
+		const packageName = args[0];
```
LamboCreeper, line 15: No need to change this, but just a tip - you can do this instead:
```js
const [packageName] = args;
```

LamboCreeper (approved):
Thanks for this!

PR #40, opened 2020-06-11
Rework InstantAnswerService

Jason2605:
# Rework InstantAnswerService
## Summary
Currently when the `InstantAnswerService` returns no results the logic is set so that an exception is thrown, and the bot returns a message similar to "there was an error". Instead if no results are returned from DDG we want it to display "no results" rather than the aforementioned message as the previous message gives off the feeling something programmatically went wrong, when in fact, it did not.

### Changes
This PR changes `InstantAnswerService` to return `null` when no results have been found. This is also changed in `SearchCommand` to handle a `null` value being returned.

### DIFF
--- a/src/services/InstantAnswerService.ts
+++ b/src/services/InstantAnswerService.ts
@@ -17,7 +17,7 @@ class InstantAnswerService {
 		return this.instance;
 	}
 
-	async query(query: string): Promise<InstantAnswer|null> {
+	async query(query: string): Promise<InstantAnswer | null> {
 		const url = `https://api.duckduckgo.com/?q=${query}&format=json&t=codesupport-discord-bot`;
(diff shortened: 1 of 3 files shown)


src/services/InstantAnswerService.ts
```diff
@@ -17,28 +17,32 @@ class InstantAnswerService {
 		return this.instance;
 	}
 
-	async query(query: string): Promise<InstantAnswer> {
+	async query(query: string): Promise<InstantAnswer|null> {
```
LamboCreeper: Can you put a space on either side of the pipe, please? Reads better in my opinion.

LamboCreeper (changes requested):

LamboCreeper (approved):

PR #44, opened 2020-06-13
Added botless channels support and wrote tests for CommandParserHandler.

### LINKED ISSUE #43: Stop Commands Being Run In #hiring-or-looking
Commands should not be allowed to run in the #hiring-or-looking channel (ID: `328146607486795777`).

ajsaraujo:
Resolves #43

### DIFF
--- a/src/event/handlers/CommandParserHandler.ts
+++ b/src/event/handlers/CommandParserHandler.ts
@@ -1,7 +1,8 @@
 import { Constants, Message } from "discord.js";
-import { COMMAND_PREFIX, BOTLESS_CHANNELS } from "../../config.json";
 import EventHandler from "../../abstracts/EventHandler";
 import CommandFactory from "../../factories/CommandFactory";
+import getConfigValue from "../../utils/getConfigValue";
+import { COMMAND_PREFIX } from "../../config.json";
 
 class CommandParserHandler extends EventHandler {
 	private readonly commandFactory: CommandFactory;
@@ -15,6 +16,8 @@ class CommandParserHandler extends EventHandler {
 
 	handle = async (message: Message): Promise<void> => {
 		if (message.content.startsWith(COMMAND_PREFIX)) {
+			const BOTLESS_CHANNELS = getConfigValue("BOTLESS_CHANNELS");
+
 			if (Object.values(BOTLESS_CHANNELS).includes(message.channel.id)) {
 				return;
 			}
--- a/src/event/handlers/CommandParserHandler.ts
+++ b/src/event/handlers/CommandParserHandler.ts
@@ -1,5 +1,5 @@
 import { Constants, Message } from "discord.js";
-import { COMMAND_PREFIX } from "../../config.json";
+import { COMMAND_PREFIX, BOTLESS_CHANNELS } from "../../config.json";
 import EventHandler from "../../abstracts/EventHandler";
 import CommandFactory from "../../factories/CommandFactory";
 
@@ -15,6 +15,10 @@ class CommandParserHandler extends EventHandler {
 
 	handle = async (message: Message): Promise<void> => {
 		if (message.content.startsWith(COMMAND_PREFIX)) {
+			if (Object.values(BOTLESS_CHANNELS).includes(message.channel.id)) {
+				return;
+			}
+
 			const args = message.content.replace("?", "").split(" ");
 			const trigger = args.shift() || args[0];
 
--- a/src/utils/getConfigValue.ts
+++ b/src/utils/getConfigValue.ts
@@ -1,0 +1,11 @@
+import config from "../config.json";
+
+function getKeyValue(key: string, obj: Record<string, any>) {
+	return obj[key];
+}
+
+function getConfigValue(key: string) {
+	return getKeyValue(key, config);
+}
+
+export default getConfigValue;
--- a/test/event/handlers/CommandParserHandlerTest.ts
+++ b/test/event/handlers/CommandParserHandlerTest.ts
@@ -3,12 +3,11 @@ import { Constants } from "discord.js";
 import { Message } from "discord.js";
 import { SinonSandbox, createSandbox } from "sinon";
 
-import { COMMAND_PREFIX, BOTLESS_CHANNELS } from "../../../src/config.json";
 import MockCommand from "../../MockCommand";
 import MockDiscord from "../../MockDiscord";
-import Command from "../../../src/abstracts/Command";
 import CommandFactory from "../../../src/factories/CommandFactory";
 import CommandParserHandler from "../../../src/event/handlers/CommandParserHandler";
+import { COMMAND_PREFIX, BOTLESS_CHANNELS } from "../../../src/config.json";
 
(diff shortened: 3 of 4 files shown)


test/event/handlers/CommandParserHandlerTest.ts
```diff
@@ -0,0 +1,128 @@
...
+			await handler.handle(message);
+
+			expect(runCommandMock.called).to.be.false;
+		});
+
+		it("should not run command if message was sent on a botless channel", async () => {
+			const runCommandMock = sandbox.stub(command, "run");
+
+			const message = discordMock.getMessage();
+			const botlessChannelId : string = Object.values(BOTLESS_CHANNELS)[0] as string;
```
LamboCreeper: Can you remove the space before the `:` please?

LamboCreeper: What happens if `BOTLESS_CHANNELS` is empty? Does this test fail?
Probably better just to make up an ID as we're not really doing integration tests anyway.

test/event/handlers/CommandParserHandlerTest.ts
```diff
@@ -0,0 +1,128 @@
...
+			const message = discordMock.getMessage();
+
+			message.content = COMMAND_PREFIX + command.getName();
+			message.channel.id = "fake-bot-enabled-channel-id-123";
+
+			await handler.handle(message);
+
+			expect(runCommandMock.called).to.be.true;
+		});
+		it("should delete messages that trigger a self destructing command", async () => {
```
LamboCreeper: Can you add a gap between these two tests, please?

test/event/handlers/CommandParserHandlerTest.ts
```diff
@@ -0,0 +1,128 @@
...
+			sandbox.stub(CommandFactory.prototype, "commandExists").callsFake(() => true);
+			sandbox.stub(CommandFactory.prototype, "getCommand").callsFake(() => command);
+			sandbox.stub(MockCommand.prototype, "isSelfDestructing").callsFake(() => true);
+			const deleteMessageMock = sandbox.stub(Message.prototype, "delete");
+
+			const message = discordMock.getMessage();
```
LamboCreeper: Can you put a gap after the final `sandbox.stub()` and move the `const`s together, please?

LamboCreeper (changes requested):
Thanks for writing those tests, I appreciate it! Just a few code-style changes and a question regarding functionality.

test/event/handlers/CommandParserHandlerTest.ts
ajsaraujo: Oops, I didn't think about that case.

The test passes, but for the wrong reason: the command isn't ran because it's not recognized by `CommandFactory`, so `commandExists()` returns `false`. In the other test cases I mock `commandExists()` so it returns `true`. I'm wondering if I should also mock `commandExists()` to return true in the prefix test to avoid this behavior.

When you say "make up an id", you mean adding a fake channel id to the `BOTLESS_CHANNELS` object in `config.json`? That seems like a good workaround to me.

LamboCreeper: Nah, don't add the fake id to the `BOTLESS_CHANNELS` object, just hard code it in the test.

ajsaraujo: If I hardcode the channel ID, this expression will evaluate to `false`, making the test fail:
`Object.values(BOTLESS_CHANNELS).includes("my-hardcoded-channel-id")`

LamboCreeper: Oh, I see - my bad. Is there a way to mock the variable? If not I guess we'll just have to do it with a value in the `config.json`.

ajsaraujo: Sadly, Sinon only let's us stub method calls, so we wouldn't be able to stub the export value of `config.json`. But apparently there's a NPM package called `proxyquire` that let's you do this (override dependencies). 

Do you think it would be interesting to use it, or should we stay with the mock channel thing? The latter seems a bit inelegant in my opinion.

https://stackoverflow.com/questions/56890285/how-to-stub-a-json-file-with-sinon

LamboCreeper: Do whatever you thinks best, I don't mind! :)

ajsaraujo (commented):
@LamboCreeper, I made the style fixes you asked. About the botless channel test, I went with the mock channel solution. It's not very nice to have mock data in the config file, but sadly it seems that the NPM package I mentioned before won't suit our needs, since it mocks the `require` function, while we are using `import`.

LamboCreeper:
> @LamboCreeper, I made the style fixes you asked. About the botless channel test, I went with the mock channel solution. It's not very nice to have mock data in the config file, but sadly it seems that the NPM package I mentioned before won't suit our needs, since it mocks the `require` function, while we are using `import`.

You should still be able to use `import`, might just have to do the `* as Variable` syntax.

src/event/handlers/CommandParserHandler.ts
```diff
@@ -1,7 +1,8 @@
 import { Constants, Message } from "discord.js";
-import { COMMAND_PREFIX } from "../../config.json";
 import EventHandler from "../../abstracts/EventHandler";
 import CommandFactory from "../../factories/CommandFactory";
```
ajsaraujo, line 3: I didn't change all `config.json` calls to `getConfigValue` calls because I didn't know if it was the right thing to do. Should we only change it for variables that need to be mocked? Or should all calls be changed?

src/event/handlers/CommandParserHandler.ts
```diff
@@ -15,6 +16,12 @@ class CommandParserHandler extends EventHandler {
 
 	handle = async (message: Message): Promise<void> => {
 		if (message.content.startsWith(COMMAND_PREFIX)) {
+			const BOTLESS_CHANNELS = getConfigValue("BOTLESS_CHANNELS");
```
ajsaraujo, line 19: I wonder if this has a worse performance than a regular import, but my Node knowledge is still incipient, please correct me if I'm wrong. I thought it would be better to move this line up along with the imports, but apparently `getConfigValue()` would be only called once during the program runtime and then the stub wouldn't cause any effect.

src/utils/getConfigValue.ts
```diff
@@ -0,0 +1,11 @@
+import config from "../config.json";
+
+function getKeyValue(key: string, obj: Record<string, any>) {
+	return obj[key];
+}
+
+function getConfigValue(key: string) {
```
ajsaraujo, line 7: Some TypeScript trickery. The interpreter wouldn't let me access `obj[key]` directly without the `getKeyValue()` wrapper.

src/event/handlers/CommandParserHandler.ts
LamboCreeper: I'm not too worried about performance right now, honestly. Should be fine as is.

LamboCreeper (approved):

PR #45, opened 2020-06-14
Support Use of .env Files

LamboCreeper:
A much-requested feature has been for us to support `.env` files, I didn't really get the point of it since we only have one, however, with the TwitterService I'm working on it will require an additional four. This will make environment variables easier to work with.

#### Overview
- Installed dotenv
- Added `.env` to the `.gitignore`
- Specified non-production environments to use `.env` file (not required, however)
- Added to README and explained where to create the file

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+DISCORD_TOKEN=exampleToken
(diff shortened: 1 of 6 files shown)


Jason2605 (approved):
Wonderful 👍

PR #48, opened 2020-06-19
Post Tweets In General

LamboCreeper:
This replaces our reliance on an external service, Zapier, to post tweets from CodeSupport into the general channel. It also allows us to filter out replies from spamming the channel.

**ESLint Rule Updates**
- `no-invalid-this` is now `off`: it appears this is to do with how TypeScript generates classes?
- `curly` is now `["error", "multi-line"]`: this allows us to not need `{}` when an if statement does not have a body

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,8 @@ module.exports = {
     },
     rules: {
         "no-unused-vars": "off",
+        "no-invalid-this": "off",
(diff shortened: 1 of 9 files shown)


Jason2605 (approved):

Jason2605:
When it comes round to deployment, remember to add the twitter credentials to the environment 👍

PR #56, opened 2020-06-23
Full text from tweet in embed

BananaCrazy:


src/interfaces/TwitterStreamListener.ts
```diff
@@ -1,6 +1,8 @@
 interface TwitterStreamListener {
 	id_str: string;
-	text: string;
+	extended_tweet: {
+		full_text: string;
+	};
```
LamboCreeper, line 5: I don't believe you need the `;` here.

test/services/TwitterServiceTest.ts
```diff
@@ -67,7 +68,9 @@ describe("TwitterService", () => {
 
 			await twitterService.handleTwitterStream({
 				id_str: "",
-				text: "@This does not start with an @"
+				extended_tweet: {
+					full_text: "@This does not start with an @"
```
LamboCreeper: This is totally my fault, but the message doesn't make sense... it does begin with an "@".

LamboCreeper (changes requested):

LamboCreeper (approved):

PR #59, opened 2020-06-26
Fix unhandled TypeError "Cannot read property 'toLowerCase' of undefined"

ajsaraujo:
This error happened when a message consisting only of a `?` (our Command prefix) was sent. After `CommandParserHandler`'s parsing, `undefined` would be passed to `CommandFactory.commandExists()`, causing it to throw this error. Just had to add an extra check inside `CommandParserHandler` to prevent this from happening.

### DIFF
--- a/src/event/handlers/CommandParserHandler.ts
+++ b/src/event/handlers/CommandParserHandler.ts
@@ -25,6 +25,8 @@ class CommandParserHandler extends EventHandler {
 			const args = message.content.replace("?", "").split(" ");
 			const trigger = args.shift() || args[0];
 
+			if (!trigger) return;
+
 			/* eslint-disable */
 			if (this.commandFactory.commandExists(trigger)) {
 				const command = this.commandFactory.getCommand(trigger);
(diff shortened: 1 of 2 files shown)


src/event/handlers/CommandParserHandler.ts
```diff
@@ -25,6 +25,8 @@ class CommandParserHandler extends EventHandler {
 			const args = message.content.replace("?", "").split(" ");
 			const trigger = args.shift() || args[0];
 
+			if (!trigger) return;
```
ajsaraujo, line 28: This check could have been done in many other ways, such as checking if message's length is bigger than 1 before continuing, so it would stop if message was `"?"`. However, I think this more consistent.

test/event/handlers/CommandParserHandlerTest.ts
```diff
@@ -92,6 +93,23 @@ describe("CommandParserHandler", () => {
...
+
+			message.content = COMMAND_PREFIX;
+			message.channel.id = "fake-bot-enabled-channel-id";
+
+			let errorWasThrown = false;
+
+			try {
+				await handler.handle(message);
+			} catch (error) {
+				errorWasThrown = true;
```
ajsaraujo, line 107: I have to admit this is not optimal. I tried using Chai's error testing API with something along these lines:
```js
await handler.handle(message);

const error = new TypeError("Cannot read property 'toLowerCase' of undefined");

expect(CommandFactory.prototype.commandExists).not.to.throw(error);
```

But I couldn't manage to make it work. If anybody knows a nicer way of doing this I'll be happy to fix it.

LamboCreeper: Yeah, Chai is rather weird about handling Promise rejections. I remember @Jason2605 and I trying to figure it out. I'm happy with this.

LamboCreeper (approved):

PR #60, opened 2020-06-26
GitHub service

BananaCrazy:


src/commands/GitHubCommand.ts
```diff
@@ -0,0 +1,52 @@
...
+class GitHubCommand extends Command {
+	constructor() {
+		super(
+			"github",
+			"Shows the repo of the given user."
+		);
+	}
+
+	async run(message: Message, args: string[]) {
+		const embed = new MessageEmbed();
```
LamboCreeper, line 15: Since no matter what happens you end up sending the embed, you might as well just do it once.

src/commands/GitHubCommand.ts
```diff
@@ -0,0 +1,52 @@
...
+			embed.setDescription("You must provide a username and repo from GitHub.");
+			embed.addField("Correct Usage", "?github <username>/<repository>");
+			embed.setColor(EMBED_COLOURS.ERROR);
+
+			await message.channel.send({ embed });
+		} else {
+			const [user, repoName] = args[0].split("/");
+
+			try {
+				const GitHubRepo = GitHubService.getInstance();
```
LamboCreeper: This variable's name isn't true - can you rename it to `GitHub` or `GitHubService` please.

src/services/GitHubService.ts
```diff
@@ -0,0 +1,56 @@
...
+			throw new Error("There was a problem with the request to GitHub.");
+		}
+	}
+
+	async getPullRequest(user: string, repo: string): Promise<GitHubPullRequest[]> {
+		const url = `https://api.github.com/repos/${user}/${repo}/pulls`;
+		const { data } = await axios.get(url);
+
+		if (data.length !== 0) {
+			const pullRequests = data.map((pull: any) => ({
```
LamboCreeper, line 43: It annoys me that you're using `any` here, although I'm not sure how to combat it without creating another interface (which I don't want).

LamboCreeper (changes requested):

src/services/GitHubService.ts
BananaCrazy: Yeah it was giving me errors in VSCode and it gave that as option (and it made it work) so I just ended up using it.

LamboCreeper (approved):
Awesome - thanks for this!

PR #63, opened 2020-06-28
GitHub args check

BananaCrazy:


LamboCreeper (changes requested):
I would add an additional test which checks if the message is in the correct format, logically it will be the same as checking if no argument is given.

LamboCreeper (approved):
Thanks :)

PR #64, opened 2020-06-29
GitHub extra info

BananaCrazy:


Jason2605 (approved):

src/commands/GitHubCommand.ts
```diff
@@ -35,6 +35,9 @@ class GitHubCommand extends Command {
 				embed.addField("Language", `${res.language}`, true);
 				embed.addField("Open issues", `${res.issues_and_pullrequests_count - resPR.length}`, true);
 				embed.addField("Open Pull Requests", `${resPR.length}`, true);
+				embed.addField("Forks", `${res.forks}`, true);
+				embed.addField("Stars", `${res.stars}`, true);
+				embed.addField("Watchers", `${res.watchers}`, true);
```
LamboCreeper: I totally missed this last time, but is there a reason these are template literals rather than just putting the variable straight in as a parameter?

BananaCrazy: Oh, probably because of a copy paste that started from the title and description, and I just used it on all the rest. Want me to change that? Looked over it myself really

LamboCreeper: Yeah, I'd change it as there's no point doing it as is. :)

LamboCreeper (approved):

PR #70, opened 2020-07-04
GitHub issues command

### LINKED ISSUE #69: GitHub Issues Command
Usage: `?issues <username>/<repository>`

```
As a: user
I want to: get a list of issues for a specific repository
So that I: can see them in Discord
```

This will require updating the `GitHubService` so that it has a new method, `getIssues()`, which returns an array of `GitHubIssue`s (this will require a new interface).

When executing the command, an embed should be sent with the 3 most recent issues:
Example Embed

BananaCrazy:


src/commands/IssuesCommand.ts
```diff
@@ -0,0 +1,63 @@
...
+			embed.setColor(EMBED_COLOURS.ERROR);
+		} else {
+			const [user, repoName] = args[0].split("/");
+
+			try {
+				const GitHub = GitHubService.getInstance();
+				const resIssues = await GitHub.getIssues(user, repoName);
+				const resRep = await GitHub.getRepository(user, repoName);
+
+				if (resIssues.length > 0) {
```
LamboCreeper: Pretty sure you can just do `if (resIssues.length)`

src/commands/IssuesCommand.ts
```diff
@@ -0,0 +1,63 @@
...
+					embed.setTitle(`GitHub Issues: ${user}/${repoName}`);
+					embed.setDescription(`${resRep.description}\n\n[View Issues on GitHub](${resRep.url}/issues) - [Create one](${resRep.url}/issues/new)`);
+
+					resIssues.forEach((issue: GitHubIssue) => {
+						const diff = Math.abs(Date.now() - issue.created_at.getTime());
+						const days = Math.ceil(diff / (1000 * 3600 * 24));
```
LamboCreeper: This could be broken into a helper `getDaysBetweenDates(date1: Date, date2: Date): number`. Could be useful in the future.

src/services/GitHubService.ts
```diff
@@ -55,6 +56,28 @@ class GitHubService {
...
 	}
+
+	async getIssues(user: string, repo: string): Promise<GitHubIssue[]> {
+		const url = `https://api.github.com/repos/${user}/${repo}/issues`;
+
+		const { data } = await axios.get(url);
+
+		if (data.length !== 0) {
+			const issues = data.filter((issueAndPr: any) => !issueAndPr.pull_request)
+				.slice(0, 3)
```
LamboCreeper: I feel as if we shouldn't do the slicing in here but in the command. Just in case we want to use this method in the future but need more than 3.

test/services/GitHubServiceTest.ts
```diff
@@ -109,4 +109,50 @@ describe("GitHubService", () => {
 			sandbox.restore();
 		});
 	});
+
+	describe("getIssues", () => {
```
LamboCreeper: Missing `()` after `getIssues`

LamboCreeper (changes requested):

LamboCreeper (approved):

PR #71, opened 2020-07-05
Coverage

LamboCreeper:
Hopefully, whenever we push to master a coverage report will be uploaded to Codacy. I'm not totally sure if this will work, but it's a bit of trial and error.


ajsaraujo (approved):
Looks alright! :+1:

PR #76, opened 2020-07-06
Updated embed message for issues

BananaCrazy:
Tests for DateUtils was supposed to be a seperate commit but hey ho

### DIFF
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -6,6 +6,7 @@ class DateUtils {
 	}
 
 	static formatDaysAgo(days: number): string {
+		if (days < 0) throw new Error("Number has to be positive");
 		if (days === 0) return "Today";
 		if (days === 1) return "Yesterday";
 		return `${days} day(s) ago`;
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -7,9 +7,9 @@ class DateUtils {
 
 	static formatDaysAgo(days: number): string {
 		if (days < 0) throw new Error("Number has to be positive");
-		if (days === 0) return "Today";
-		if (days === 1) return "Yesterday";
-		return `${days} day(s) ago`;
+		if (days === 0) return "today";
+		if (days === 1) return "yesterday";
+		return `${days} days ago`;
 	}
 }
 
--- a/src/utils/DateUtils.ts
+++ b/src/utils/DateUtils.ts
@@ -1,0 +1,15 @@
+class DateUtils {
+	static getDaysBetweenDates(date1: Date, date2: Date): number {
+		const diff = date1.getTime() - date2.getTime();
+
+		return Math.floor(diff / (1000 * 3600 * 24));
+	}
+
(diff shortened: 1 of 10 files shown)


src/utils/DateUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+		const diff = date1.getTime() - date2.getTime();
+
+		return Math.floor(diff / (1000 * 3600 * 24));
+	}
+
+	static formatDaysAgo(days: number): string {
+		if (days < 0) throw new Error("Number has to be positive");
+		if (days === 0) return "Today";
+		if (days === 1) return "Yesterday";
+		return `${days} day(s) ago`;
```
LamboCreeper: The `(s)` isn't needed - there is no way it can be a singular day.

src/utils/DateUtils.ts
```diff
@@ -0,0 +1,16 @@
...
+	static getDaysBetweenDates(date1: Date, date2: Date): number {
+		const diff = date1.getTime() - date2.getTime();
+
+		return Math.floor(diff / (1000 * 3600 * 24));
+	}
+
+	static formatDaysAgo(days: number): string {
+		if (days < 0) throw new Error("Number has to be positive");
+		if (days === 0) return "Today";
+		if (days === 1) return "Yesterday";
```
LamboCreeper: Can these be in lowercase, please? I feel as if a util shouldn't have any capitalisation.

test/utils/DateUtilsTest.ts
```diff
@@ -0,0 +1,47 @@
+import { createSandbox, SinonSandbox } from "sinon";
+import { expect } from "chai";
+import DateUtils from "../../src/utils/DateUtils";
+
+describe("DateUtils", () => {
+	describe("getDaysBetweenDates()", () => {
```
LamboCreeper: Since this is a static method can you add `::` before the method name?

test/utils/DateUtilsTest.ts
```diff
@@ -0,0 +1,47 @@
...
+		});
+
+		it("returns 1 if it's a day later", () => {
+			const days = DateUtils.getDaysBetweenDates(new Date("2020-01-02"), new Date("2020-01-01"));
+
+			expect(days).to.be.equal(1);
+		});
+	});
+
+	describe("getDaysBetweenDates()", () => {
```
LamboCreeper: Since this is a static method can you add `::` before the method name?

LamboCreeper (changes requested):

src/utils/DateUtils.ts
BananaCrazy: My thinking was if it was needed at the start of a sentence it could be used like this, in other cases it would be easy to add `.toLowerCase()`.

LamboCreeper: I guess, but it's sort of assuming use-case when utils are meant to be standalone.

LamboCreeper (approved):

LamboCreeper (approved):

PR #82, opened 2020-07-09
Twitter retweets

### LINKED ISSUE #81: Don't Post Retweets
Currently, when someone retweets any tweet made by us it sends it in general with the prefix ”RT”.

This shouldn't happen as it gets quite spammy.

Although, it should still post any tweets that we retweet.

BananaCrazy:


src/services/TwitterService.ts
```diff
@@ -31,12 +31,14 @@ class TwitterService {
 		}).on("data", listener => this.handleTwitterStream(listener, tweetChannel));
 	}
 
-	handleTwitterStream = async ({ id_str: id, text, extended_tweet }: TwitterStreamListener, tweetChannel: TextChannel): Promise<void> => {
+	handleTwitterStream = async ({ id_str: id, text, user, retweeted_status, extended_tweet }: TwitterStreamListener, tweetChannel: TextChannel): Promise<void> => {
```
LamboCreeper: This line is starting to get rather long, can we do the destructing on a new line, i.e.:
```ts
handleTwitterStream = async (stream: TwitterStreamListener, tweetChannel: TextChannel): Promise<Void> => {
    const { id_str: id, text, user, retweeted_status, extended_tweet} = stream;
    ...
}
```

src/services/TwitterService.ts
```diff
@@ -31,12 +31,14 @@ class TwitterService {
...
 		const embed = new MessageEmbed();
 		const tweet = extended_tweet?.full_text || text;
+		const tweetId = retweeted_status?.id_str || id;
+		const retweetFromCodeSupport = retweeted_status && user?.id_str === TWITTER_ID ? true : false;
```
LamboCreeper: Do you need `? true : false`? Surely `user?.id_str === TWITTER_ID` results in a `Boolean`?

LamboCreeper (changes requested):
Looks good! Just some style questions to make it more readable.

LamboCreeper (approved):

PR #92, opened 2020-07-25
Raid Detection

saramaebee:


saramaebee:
Note: while the test fails on CI, that's because it takes longer than twenty seconds. Test passes on computer, can be verified by lowering the `TIME_TILL_REMOVAL` option in config.json

src/config.json
```diff
@@ -1,6 +1,7 @@
 {
     "COMMAND_PREFIX": "?",
-    "MEMBER_ROLE": "592088198746996768",
+	"MEMBER_ROLE": "592088198746996768",
+	"MOD_ROLE":"490594428549857281",
```
LamboCreeper: Formatting is not consistent here.

src/config.json
```diff
@@ -11,12 +12,17 @@
     "PRODUCTION_ENV": "production",
     "commands_directory": "commands",
     "handlers_directory": "event/handlers",
-    "LOG_CHANNEL_ID": "405068878151024640",
+	"LOG_CHANNEL_ID": "405068878151024640",
+	"MODS_CHANNEL_ID": "495713774205141022",
```
LamboCreeper: Formatting is not consistent here.

src/event/handlers/RaidDetectionHandler.ts
```diff
@@ -0,0 +1,33 @@
+import { Constants, GuildMember, TextChannel } from "discord.js";
+import EventHandler from "../../abstracts/EventHandler";
+import { MODS_CHANNEL_ID, RAID_SETTINGS, MOD_ROLE } from "../../config.json";
+
+class RaidDetectionHandler extends EventHandler {
+	constructor() {
+		super(Constants.Events.GUILD_MEMBER_ADD);
+	}
+
+	private joinQueue: GuildMember[] = [];
```
LamboCreeper: Can this be above the constructor please?

src/event/handlers/RaidDetectionHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+	handle = async (member: GuildMember): Promise<void> => {
+		const timeToWait = 1000 * RAID_SETTINGS.TIME_TILL_REMOVAL;
+		const modChannel = member.guild?.channels.cache.find(channel => channel.id === MODS_CHANNEL_ID) as TextChannel;
+
+		this.joinQueue.push(member);
+
+		if (this.joinQueue.length > RAID_SETTINGS.MAX_QUEUE_SIZE) {
+			await modChannel?.send(`<@&${MOD_ROLE}>, a raid has been detected!`);
+		}
+		setTimeout(() => {
```
LamboCreeper: Can you add a blank line between these two, please.

src/event/handlers/RaidDetectionHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+			const index = this.joinQueue.indexOf(member);
+
+			if (index > -1) {
```
LamboCreeper, line 26: Could you not just do this?
```suggestion
			if (this.joinQueue.includes(member)) {
```

test/MockDiscord.ts
```diff
@@ -59,8 +59,8 @@ export default class MockDiscord {
 		return this.user;
 	}
 
-	public getGuildMember(): Discord.GuildMember {
-		return this.guildMember;
+	public getGuildMember(unique?:boolean): Discord.GuildMember {
```
LamboCreeper: Can you add a space after the `:` please.

test/event/handlers/RaidDetectionHandlerTest.ts
```diff
@@ -0,0 +1,64 @@
...
+		it("removes member from joinQueue", done => {
+			const mockGuildMember = discordMock.getGuildMember();
+
+			handler.handle(mockGuildMember).then(() => {
+				expect(handler.joinQueue.includes(mockGuildMember)).to.be.true;
+				setTimeout(() => {
+					expect(handler.joinQueue.includes(mockGuildMember)).to.be.false;
+					done();
+				}, 1000 + RAID_SETTINGS.TIME_TILL_REMOVAL * 1000);
+			});
```
LamboCreeper: Is there a reason you are using `.then()` over `async/await` here?

src/event/handlers/RaidDetectionHandler.ts
```diff
@@ -0,0 +1,33 @@
...
+				this.joinQueue.splice(index, 1);
+			} else {
+				modChannel?.send("Had trouble removing user from join queue");
```
LamboCreeper: Might be good to specify which user?

LamboCreeper (changes requested):
Looks good, just a few formatting things. We're going to need to fix the ability for this not to run in CI before we merge, though.

src/event/handlers/RaidDetectionHandler.ts
saramaebee: i need the index to splice it, don't i? I suppose i could do 
```js
if (this.joinQueue.includes(member)) {
    this.joinQueue.splice(this.joinQueue.indexOf(member), 1);
}
```

test/event/handlers/RaidDetectionHandlerTest.ts
saramaebee: that was the only way to make it work. see discord bot channel

saramaebee:
> Looks good, just a few formatting things. We're going to need to fix the ability for this not to run in CI before we merge, though.

i mean i can set the `TIME_TILL_REMOVAL` to be less than 20, youll just have to put it back up to a higher number on the prod version each time.

src/event/handlers/RaidDetectionHandler.ts
LamboCreeper: Oh right, my bad - yeah, just leave it as it was then! 👍

LamboCreeper:
> > Looks good, just a few formatting things. We're going to need to fix the ability for this not to run in CI before we merge, though.
> 
> i mean i can set the `TIME_TILL_REMOVAL` to be less than 20, youll just have to put it back up to a higher number on the prod version each time.

Hm, can we mock the value? We can't have a different config on prod to what it is in CI.

saramaebee:
> > > Looks good, just a few formatting things. We're going to need to fix the ability for this not to run in CI before we merge, though.
> > 
> > 
> > i mean i can set the `TIME_TILL_REMOVAL` to be less than 20, youll just have to put it back up to a higher number on the prod version each time.
> 
> Hm, can we mock the value? We can't have a different config on prod to what it is in CI.

to mock the value, i could probably do some fanangling in the handle method with an optional boolean value `test`, and set it to only hold for five seconds if true?

LamboCreeper:
> > > > Looks good, just a few formatting things. We're going to need to fix the ability for this not to run in CI before we merge, though.
> > > 
> > > 
> > > i mean i can set the `TIME_TILL_REMOVAL` to be less than 20, youll just have to put it back up to a higher number on the prod version each time.
> > 
> > 
> > Hm, can we mock the value? We can't have a different config on prod to what it is in CI.
> 
> to mock the value, i could probably do some fanangling in the handle method with an optional boolean value `test`, and set it to only hold for five seconds if true?

Oh, it's easier then that - we built a [getConfigValue()](https://github.com/codesupport/discord-bot/blob/develop/src/utils/getConfigValue.ts) method to do this - you can call that instead of `config.MY_VALUE` and easily mock it's value in a test. :)

saramaebee:
All changes resolved

LamboCreeper (approved):

PR #96, opened 2020-07-30
Write tests for app, getEnvironmentVariable() and CommandFactory.loadCommands()

ajsaraujo:


test/factories/CommandFactoryTest.ts
```diff
@@ -17,12 +19,53 @@ describe("CommandFactory", () => {
 		aliases = new MockCommandWithAlias().getAliases();
 
 		factory = new CommandFactory();
-		factory["commands"] = {};
-		factory["commands"][commandName] = () => new MockCommand();
-		factory["commands"][commandWithAliasesName] = () => new MockCommandWithAlias();
+		factory.commands = {};
+		factory.commands[commandName] = () => new MockCommand();
+		factory.commands[commandWithAliasesName] = () => new MockCommandWithAlias();
```
ajsaraujo, line 25: Oops. Only realized this now, I must have done it while using "Fix all auto-fixable problems". Let me know if you want me to undo it

LamboCreeper: Nah, `factory.commands` makes more sense then `factory["commands"]`.

LamboCreeper (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "The CodeSupport Discord bot",
   "main": "./build/app.js",
   "scripts": {
-    "start": "node -r dotenv/config ./build/app.js",
+    "start": "node -r dotenv/config ./build/index.js",
```
ajsaraujo, line 7: As I discussed with @Jason2605 in #discord-bot, `index.ts` was created to separate app definition and execution. If everything is in the same place, the app is going to be executed when a test imports it, which we don't want to happen.

test/app/appTest.ts
```diff
@@ -0,0 +1,97 @@
+import { expect } from "chai";
```
LamboCreeper: Why is this in a folder called `app`? That doesn't match `src/`'s structure.

LamboCreeper (approved):
Amazing. Thank you so much!

test/app/appTest.ts
ajsaraujo: When I put inside `test/`, the tests weren't ran. But now I realized that I had to tweak the test script, already fixed it. :+1:

LamboCreeper (approved):

PR #104, opened 2020-08-07
Ghost Ping Detection

jacobgasser:
When a user deletes a message that contains a ping to a user or a role, an embed will be sent alerting the user that was pinged.

This feature will stop users from scrolling looking for pings and users can also search for messages that mention them even if the message was deleted.

### DIFF
--- a/src/event/handlers/GhostPingMessageDeleteHandler.ts
+++ b/src/event/handlers/GhostPingMessageDeleteHandler.ts
@@ -1,0 +1,24 @@
+import { Constants, MessageEmbed, Message } from "discord.js";
+import { EMBED_COLOURS } from "../../config.json";
+import EventHandler from "../../abstracts/EventHandler";
+
+class GhostPingMessageDeleteHandler extends EventHandler {
+	constructor() {
+		super(Constants.Events.MESSAGE_DELETE);
+	}
+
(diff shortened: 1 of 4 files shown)


src/event/handlers/GhostPingMessageDeleteHandler.ts
```diff
@@ -0,0 +1,24 @@
+import { Constants, MessageEmbed, Message } from "discord.js";
+import { EMBED_COLOURS } from "../../config.json";
+import EventHandler from "../../abstracts/EventHandler";
+
+class GhostPingMessageDeleteHandler extends EventHandler {
```
ajsaraujo: I think `GhostPingHandler` is more coherent with the stablished convention, besides being more concise. Anyways, good job Jacob :wink:

jacobgasser: I can fix it

ajsaraujo (approved):
Looking good! :+1:

LamboCreeper:
Looks good, but before I approve we need to think about an edge case: what happens if a user is banned or has their messages deleted by a Moderator or spam detection bot? Would the chat get spammed with these embeds? Would nothing happen?

LamboCreeper (approved):

PR #113, opened 2020-09-11
Improve quote system

### LINKED ISSUE #97: Quote System improvements
- Serialize URLs
- Display date and time as depicted [here](https://cdn.discordapp.com/attachments/720675405413089311/738315596927598632/unknown.png)
- Display which channel the message was sent in

ajsaraujo:
I made minor refactorings in `MessagePreviewService` and `MessagePreviewServiceTest`, besides implementing the requested features. Below there's a screenshot of how the quote system is looking like right now:

image
Resolves #97


src/config.json
```diff
@@ -9,6 +9,8 @@
         "html",
         "css"
     ],
+    "FIELD_SPACER_CHAR": "\u200B",
+    "MEMBER_ROLE_COLOR": "FFFFFE",
```
LamboCreeper: This should really have a `#` for consistency.

src/services/MessagePreviewService.ts
```diff
@@ -23,31 +25,59 @@ class MessagePreviewService {
...
+					embed.setFooter(`Message sent at ${DateUtils.format(messageToPreview.createdAt)}`);
+					embed.setColor(messageToPreview.member?.displayColor || MEMBER_ROLE_COLOR);
 
 					callingMessage.channel.send(embed);
 				}
 			}
 		}
 	}
 
+	getChannelName(message: Message): String {
```
LamboCreeper: Should be `string` (you’re returning the type not a the class).

src/services/MessagePreviewService.ts
```diff
@@ -23,31 +25,59 @@ class MessagePreviewService {
...
+		const textChannel = message.channel as TextChannel;
+
+		return textChannel.name;
+	}
+
+	escapeHyperlinks(content: String): String {
```
LamboCreeper: Same here.

src/services/MessagePreviewService.ts
```diff
@@ -23,31 +25,59 @@ class MessagePreviewService {
...
+		if (!content) {
+			return content;
```
LamboCreeper: How can you return something that doesn't exist? A bit confused what's going on here.

src/utils/DateUtils.ts
```diff
@@ -11,6 +11,15 @@ class DateUtils {
 		if (days === 1) return "yesterday";
 		return `${days} days ago`;
 	}
+
+	static format(date: Date) {
```
LamboCreeper: I feel like ”format” is quite a generic name, maybe something like formatAsText?

src/services/MessagePreviewService.ts
ajsaraujo: It's supposed to return `null`, `undefined` or `""` if content matches any of those, to prevent a `TypeError` when calling `content.replace()`. Maybe an elvis operator would be better? e.g `content?.replace()`

src/services/MessagePreviewService.ts
```diff
@@ -23,31 +25,55 @@ class MessagePreviewService {
...
+					output = `${output.slice(0, i)}\\${output.slice(i)}`;
+					i++;
+				}
+			}
+
+			return output;
+		});
+	}
+
+	getAuthorName(message: Message) {
```
LamboCreeper: Missing return type

src/utils/DateUtils.ts
```diff
@@ -11,6 +11,15 @@ class DateUtils {
 		if (days === 1) return "yesterday";
 		return `${days} days ago`;
 	}
+
+	static formatAsText(date: Date) {
```
LamboCreeper: Missing return type

LamboCreeper (approved):

PR #116, opened 2020-09-19
?commands Command

### LINKED ISSUE #115: Add a commands command
It's not clear for newcomers how the bot is used - right now, the learning curve consists basically in watching other people use it. It would be nice if we had a `?commands` command that shows what are the available commands and what each one of them does. 

The command should loop over all the commands, getting their names, descriptions and aliases (if there are any) and finally showing this data on an embed.

A `help` alias would also fit nicely.

LamboCreeper:
Resolves #115 

#### Overview
- Added `?commands` command
- Added new `GenericObject<T>` type to stop the use of `any`


src/commands/CommandListCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	setCommands(commands: Command[]) {
+		this.commands = commands;
+	}
+
+	async run(message: Message, args: string[]) {
+		const embed = new MessageEmbed();
+
+		if (args.length) {
+			const command = this.commands.find(command => command.getName() === args[0]);
```
ajsaraujo: Should we also search by alias? E.g
```js
if (args.length) {
    const command = this.commands.find(command => command.getName() === args[0] || command.getAliases().includes(args[0]));
}
```

src/event/handlers/CommandParserHandler.ts
```diff
@@ -31,6 +32,10 @@ class CommandParserHandler extends EventHandler {
 			if (this.commandFactory.commandExists(trigger)) {
 				const command = this.commandFactory.getCommand(trigger);
 
+				if (command instanceof CommandListCommand) {
+					command.setCommands(this.commandFactory.getCommandsWithoutAliases());
+				}
```
ajsaraujo, line 37: I know performance isn't exactly an issue, but isn't that going to run everytime `?commands` is triggered? I mean, we could store the list as a static property (set in `CommandParserHandler` constructor) rather than have to query it everytime.

src/commands/CommandListCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+				embed.setColor(EMBED_COLOURS.DEFAULT);
+				embed.setDescription(command.getDescription());
+
+				if (aliases.length) embed.addField("Aliases", aliases.join(", "));
+			}
+		} else {
+			embed.setTitle("Commands");
+			embed.setColor(EMBED_COLOURS.DEFAULT);
+
+			this.commands.forEach(command => {
```
Jason2605, line 48: Embeds have a maximum of 25 fields. Obviously we're not close to that on command count at the moment, but it's something to be aware of eventually.

src/factories/CommandFactory.ts
```diff
@@ -20,7 +22,7 @@ class CommandFactory {
 			const aliases = new Command().getAliases();
 
 			aliases.forEach((alias: string) => {
-				this.commands[alias] = () => new Command();
+				this.commandAliases[alias] = () => new Command();
```
Jason2605, line 25: I'm not entirely sure why this change was introduced, but it has broken the activation of commands through aliases. Is this still required? Seems as though `commandAliases` is not used anywhere

src/commands/CommandListCommand.ts
ajsaraujo: When we get to that point, maybe we could send 2 embeds?

src/factories/CommandFactory.ts
LamboCreeper: I did this as I wanted to have two separate arrays: just commands and commands + aliases. This means that the embed (or any future use cases) does not need to do the filtering. I see what you mean about aliases no longer being triggered, though - will fix.

src/commands/CommandListCommand.ts
LamboCreeper: I'll create an issue to paginate, I think it's out of this scope for now (plus something for Hacktoberfest!).

src/event/handlers/CommandParserHandler.ts
LamboCreeper: I think it's fine for now honestly, I mean we create a new `CommandFactory` every time a message is sent... 😬

src/factories/CommandFactory.ts
```diff
@@ -20,17 +22,27 @@ class CommandFactory {
...
 			});
 		});
 	}
 
-	commandExists(command: string): boolean {
-		return typeof this.commands[command.toLowerCase()] !== "undefined";
+	commandOrAliasExists(command: string): boolean {
+		const commandsAndAliases = {...this.commandAliases, ...this.commands};
+
+		return typeof commandsAndAliases[command.toLowerCase()] !== "undefined";
```
ajsaraujo, line 33: Why check `typeof something !== "undefined"` instead of just `something !== undefined`?

LamboCreeper:
Closing as this feature will become irrelevant with the new Discord Interactions API.

PR #122, opened 2020-10-28
Support GUILD_MEMBERS intent.

LamboCreeper:
also made allan codeowner as i thought he already was


ajsaraujo (approved):
Looking good! The new codeowner too.

PR #128, opened 2020-11-02
Added selfping bugfix and added tests

BlackBearFTW:
- Now ghostping wont be activated when a user only mentions himself

- Added test to check if message wont get send if user only mentions himself

- Added test to check if message will be send if message author and another user will be mentioned

**NOTE: this doesnt use Lambo's new package**


BlackBearFTW:
Wait for the merge I still have to implement Lambo's mockDiscord package instead of how im currently creating users myself

PR #129, opened 2020-11-02
Bugfix/ghost self ping

### LINKED ISSUE #119: Ghost ping detection triggers even if the user only mentions themselves
See below for an example, where he mentioned himself, then later deleted the post.

image

BlackBearFTW:
- Now when a user only pings himself the ghost ping alert wont trigger, except when the user pings himself and someone else 

**NOTE TESTS STILL MANUALLY CREATE USER**
I still have to implement Lambo's new mockDiscord package

- Added test to check if a message wont get send when a user only pings himself.
- Added test to check if a message does get send when a user pings himself and someone else
- Fixed "sends a message when a message is deleted that pinged a user" test


BlackBearFTW:
Oh Yeah, this is done now and ready for review, 
Thank you Jason, for helping me with my tests

test/event/handlers/CodeblocksOverFileUploadsHandlerTest.ts
```diff
@@ -24,11 +24,12 @@ describe("CodeblocksOverFileUploadsHandler", () => {
 		beforeEach(() => {
 			sandbox = createSandbox();
 			handler = new CodeblocksOverFileUploadsHandler();
-			message = BaseMocks.getMessage();
-			message.id = "1234";
-			message.attachments = new Collection<string, MessageAttachment>();
-			message.author = BaseMocks.getUser();
+			message = CustomMocks.getMessage({
+				id: "1234",
+				author: BaseMocks.getUser(),
```
LamboCreeper: Can you remove the trailing comma please.

src/event/handlers/GhostPingHandler.ts
```diff
@@ -10,6 +10,7 @@ class GhostPingHandler extends EventHandler {
 	async handle(message: Message): Promise<void> {
 		if (message.mentions.users.first() || message.mentions.roles.first()) {
 			if (!message.author?.bot) {
+				if (message.mentions.users.first()?.id === message.author.id && message.mentions.users.size === 1) return;
```
LamboCreeper: Would be nice to shorten this line:
```suggestion
				const usersMentioned = message.mentions.users;
				
				if (usersMentioned.first()?.id === message.author.id && usersMentioned.size === 1) return;

```

LamboCreeper (changes requested):

LamboCreeper (approved):

BlackBearFTW:
Done :)

PR #132, opened 2020-11-15
Fix Raid Detection

LamboCreeper:
#### Overview
- Fixed raid detection handler
- Changed handler to kick members if detected as being apart of a raid
- Added new `GenericObject` to stop the use of `any`
- Fixed `getConfigValue` so it can support nested keys

Shoutout everyone who helped test this, you're amazing.


ajsaraujo (approved):
Looks good! 👍

PR #134, opened 2020-11-22
Mock Library

Jason2605:
# Mock Library
## Summary
This PR removes the older Discord mocking library in place for the [newly created one](https://www.npmjs.com/package/@lambocreeper/mock-discord.js).

**Note**: This PR also updates [CodeblocksOverFileUploadsHandler.ts](https://github.com/codesupport/discord-bot/blob/d81b55f83935aa69b356ed43caf0e61e6d459bee/src/event/handlers/CodeblocksOverFileUploadsHandler.ts) as without it the bot would fail to build. The changes here ensure that the `fileExtension` variable always has a string value (previously it could be undefined), and if for whatever reason the value is `""` (no extension), then this would trigger the bot to remove the file upload.


package.json
```diff
@@ -27,29 +27,30 @@
   },
   "homepage": "https://github.com/codesupport/discord-bot#readme",
   "dependencies": {
+    "@lambocreeper/mock-discord.js": "^1.0.9",
```
LamboCreeper: This should be a development dependency, only used for tests.

LamboCreeper (approved):

PR #139, opened 2020-12-05
Added querying by user and year for Advent of Code command

JonaVDM:
Added position numbering.
Added a check so that before November you get back the scores from last year.
Added a optional parameter to change the year.
Moved year out of config, now able to calculate it himself
Added a optional parameter to get a single user and see his score on the leaderboard
Added a config variable to set the amount of player shown in the leaderboard

In action:

image
image


src/commands/AdventofcodeCommand.ts
```diff
@@ -46,29 +46,78 @@ class AdventOfCodeCommand extends Command {
...
-		const link = `https://adventofcode.com/${ADVENT_OF_CODE_YEAR}/leaderboard/private/view/${ADVENT_OF_CODE_LEADERBOARD}`;
+	getYear() {
+		const date = new Date();
+
+		if (date.getMonth() > 10) {
+			return date.getFullYear();
+		}
+
+		return date.getFullYear() - 1;
+	}
```
LamboCreeper: I think this makes more sense to live in `utils/DateUtils.ts`, I can see it being reused.

src/commands/AdventofcodeCommand.ts
```diff
@@ -46,29 +46,78 @@ class AdventOfCodeCommand extends Command {
...
 		const embed = new MessageEmbed();
+		const adventOfCodeService = AdventOfCodeService.getInstance();
+
+		if (queriedYear && queriedYear >= 2015 && queriedYear <= year) {
+			year = queriedYear;
+		} else if (queriedYear) {
+			embed.setTitle("Error");
+			embed.setDescription(`Year requested not available.\nPlease query a year between 2015 and ${year}`);
+			embed.setColor(EMBED_COLOURS.ERROR);
+			message.channel.send({ embed });
```
LamboCreeper: `send()` returns a `Promise`, so would be a good idea to `await` this.

src/services/AdventOfCodeService.ts
```diff
@@ -47,4 +47,11 @@ export default class AdventOfCodeService {
 
 		return members;
 	}
+
+	async getSingelPlayer(leaderBoard: string, year: number, name: string): Promise<[number, AOCMember]> {
```
LamboCreeper: Do you mean single?

LamboCreeper (changes requested):
Awesome! Just a few minor things.

src/commands/AdventofcodeCommand.ts
JonaVDM: How would you suggest calling such a function, I can do "getYearBasedOnMonth", but that will most likely get more questions that answers.

LamboCreeper: Ah, ignore this actually. It is quite specific. I misread the logic.

JonaVDM:
Woeps forgot to rename tests!

LamboCreeper (approved):

PR #140, opened 2020-12-10
Inspect Command

### LINKED ISSUE #130: Member Inspect Command
**As a** moderator
**I want to** be able to easily gather information about a user
**So that I** can fill out the member report form quicker

--- 

**Command Usage:** `?inspect <username | discord id>`
**Executeable by:** Everyone
**Executeable in:** #bot-chat

Returns an embed with the following fields:
- ID
- Username + Discriminator (formatted as Username#Discriminator)
- Joined Date + Time
- Nickname
- Roles

The embed colour should match the colour of their most senior role to be consistent with the quote embed.

BlackBearFTW:
?inspect command

Tests are still WIP

Do not review yet


src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
+import {Message, MessageEmbed} from "discord.js";
```
LamboCreeper: Can you pad all your destructures throughout this, please? For example:
```suggestion
import { Message, MessageEmbed } from "discord.js";
```

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+			"Show information about a given user"
+		);
+	}
+
+	async run(message: Message, args: string[]) {
+		const embed = new MessageEmbed();
+
+		try {
+			let userObj;
+			let userList;
```
LamboCreeper: Can these have types since they're being initialised with no value?

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+
+			if (args.length > 0) {
+				if ((/^[0-9]+$/g).test(args[0])) {
+					// - args[0] only contains numbers so its a user ID
+					userObj = await message.guild?.members?.fetch(args[0]);
+				} else {
+					// If args[0] does not match username#0000 throw error
+					if (!(/^.*#[0-9]{4}$/).test(args[0])) throw "";
```
LamboCreeper: Can you not throw an empty error message.

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+			} else {
+				embed.setTitle("Error");
+				embed.setDescription("No user was found with this username/userID.");
+				embed.setColor(EMBED_COLOURS.ERROR);
+			}
+		} catch (error) {
+			embed.setTitle("Error");
+			embed.setDescription("Incorrect usage of command");
+			embed.addField("Correct Usage", "?inspect [username + discriminator / userID]");
+			embed.setColor(EMBED_COLOURS.ERROR);
```
LamboCreeper: Can this be `?inspect [username|id]`

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+				embed.addField("Username", userObj?.user.username);
+				embed.addField("Discriminator", userObj?.user.discriminator);
+				if (userObj?.nickname !== null) embed.addField("Nickname", userObj?.nickname);
+				if (userObj?.joinedAt !== null) embed.addField("Joined At", DateUtils.formatAsText(userObj?.joinedAt));
+				embed.addField("Roles", `${userObj.roles.cache.filter(role => role.id !== message?.guild!.id).map(role => ` ${role.toString()}`)}`);
+				embed.setColor(EMBED_COLOURS.SUCCESS);
+			} else {
+				embed.setTitle("Error");
+				embed.setDescription("No user was found with this username/userID.");
+				embed.setColor(EMBED_COLOURS.ERROR);
```
LamboCreeper: Can this be `No match found.` and then add the correct usage field beneath.

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,77 @@
...
+					// - args[0] only contains numbers so its a user ID
+					userObj = await message.guild?.members?.fetch(args[0]);
+				} else {
+					// If args[0] does not match username#0000 throw error
+					if (!(/^.*#[0-9]{4}$/).test(args[0])) throw "";
+
+					const userData = args[0].split("#");
+
+					const username = userData[0];
+					const discriminator = userData[1];
```
LamboCreeper: Can be simplified to:
```suggestion
					const [username, discriminator] = args[0].split("#");
```

LamboCreeper (changes requested):

src/commands/InspectCommand.ts
JonaVDM: Maybe I am wrong or there is plan with it in the future, but I don't see userList being used outside of the scope where it is given a value. That could be initialized where the value is given.

test/commands/InspectCommandTest.ts
```diff
@@ -0,0 +1,181 @@
...
+			expect(embed.hexColor).to.equal(EMBED_COLOURS.ERROR.toLowerCase());
+		});
+
+		it("sends a message with information if the argument was a username", async () => {
+			const messageMock = sandbox.stub(message.channel, "send");
+			const member = BaseMocks.getGuildMember();
+
+			sandbox.stub(GuildMemberManager.prototype, "fetch").resolves(new Collection([["12345", member]]));
+
+			sandbox.stub(GuildMemberRoleManager.prototype, "cache").get(() => new Collection([["12345", new Role(BaseMocks.getClient(), {
```
JonaVDM: You define this exact same role 3 times, maybe you can put it in a variable to not repeat yourself?

BlackBearFTW:
Everything has been modified and it should be all done now 👍

Jason2605:
Couple of points from myself:
- `getMemberUtil` follows the convention of other utils in that directory however it's following it as though you were exporting a single function rather than a class. I think a better name would be `DiscordUtils` for the class to follow the convention of other classes in that directory.
- `return guild.members?.fetch(value) || undefined;` in `getGuildMember` isn't really doing what you expect. Fetching by a user ID will return `Promise<GuildMember>;` if, however, you pass it an ID that does not exist an exception will be thrown rather than it returning undefined. (See error below)
    - Due to this, passing an invalid ID will result to `Incorrect usage of command` rather than `No match found.`
- Similar `.find()` will already return `undefined` if it matches nothing `|| undefined` is redundant.
- You shouldn't need to await `.find` since it's not returning a promise.
- There's a lot of optional chaining in `getGuildMember` are you sure it's required?
- I feel the run command in `InspectCommand` has a lot going on, both in terms of error handling and embed creation. I think a better way to reason about this is to have your 3 separate paths of execution (essentially the paths leading to each embed) be separated out, leaving the `run()` function itself to be quite simple to reason about.


```js
DiscordAPIError: Unknown User
  ...
  method: 'get',
  path: '/guilds/706487659748392972/members/15761643794123980',
  code: 10013,
  httpStatus: 404
}
```

BlackBearFTW:
@Jason2605 I updated the DiscordUtil, Im still working on the inspectCommand

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+	private buildInspectEmbed(memberObj: GuildMember): MessageEmbed {
+		const embed = new MessageEmbed();
+
+		embed.setTitle(`Inspecting ${memberObj?.user.tag}`);
+		embed.setColor(memberObj?.displayColor || EMBED_COLOURS.DEFAULT);
+		embed.setThumbnail(memberObj?.user.displayAvatarURL());
+		embed.addField("User ID", memberObj?.user.id);
+		embed.addField("Username", memberObj?.user.tag);
+		if (memberObj?.nickname !== null) embed.addField("Nickname", memberObj?.nickname);
+		if (memberObj?.nickname !== null) embed.addField("Joined At", DateUtils.formatAsText(memberObj?.joinedAt!));
```
JonaVDM: This is questionable, if nickname exists show the join date?
Also the ? in here are not needed, as you are now sure that `memeberObj` exists, this is for the whole buildInspectEmbed function

So it could be
```ts
if (memberObj.joinedAt) embed.addField("Joined At", DateUtils.formatAsText(memberObj.joinedAt));
```

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,58 @@
+import {GuildMember, Message, MessageEmbed} from "discord.js";
```
JonaVDM: Make sure that you use spaces {  }

src/utils/DiscordUtil.ts
```diff
@@ -0,0 +1,31 @@
+import {Guild, GuildMember} from "discord.js";
```
JonaVDM: Make sure that you use spaces {  }

test/utils/DiscordUtil.ts
```diff
@@ -0,0 +1,41 @@
+import { expect } from "chai";
+import { createSandbox, SinonSandbox } from "sinon";
+import DiscordUtil from "../../src/utils/DiscordUtil";
+import {BaseMocks, CustomMocks} from "@lambocreeper/mock-discord.js";
+import {Collection, GuildMemberManager} from "discord.js";
```
JonaVDM: Make sure that you use spaces {  }

test/commands/InspectCommandTest.ts
```diff
@@ -0,0 +1,152 @@
+import { createSandbox, SinonSandbox } from "sinon";
+import { expect } from "chai";
+import { Collection, EmbedField, GuildMember, GuildMemberRoleManager, Message, Role } from "discord.js";
+import {BaseMocks, CustomMocks} from "@lambocreeper/mock-discord.js";
```
JonaVDM: Make sure that you use spaces {  }
It also seems like CustomMocks is not being used

test/utils/DiscordUtilTest.ts
```diff
@@ -0,0 +1,41 @@
+import { expect } from "chai";
+import { createSandbox, SinonSandbox } from "sinon";
+import DiscordUtil from "../../src/utils/DiscordUtil";
+import {BaseMocks, CustomMocks} from "@lambocreeper/mock-discord.js";
+import {Collection, GuildMemberManager} from "discord.js";
```
JonaVDM: Make sure that you use spaces {  }

JonaVDM (changes requested):

src/commands/InspectCommand.ts
BlackBearFTW: Fixed

src/utils/DiscordUtil.ts
```diff
@@ -0,0 +1,31 @@
+import { Guild, GuildMember } from "discord.js";
+
+class DiscordUtil {
```
LamboCreeper: Can this be `DiscordUtils` so the names are consistent, please?

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	private buildNoMatchEmbed(): MessageEmbed {
+		const embed = new MessageEmbed();
+
+		embed.setTitle("Error");
+		embed.setDescription("No match found.");
+		embed.addField("Correct Usage", "?inspect [username|userID]");
+		embed.setColor(EMBED_COLOURS.ERROR);
+		return embed;
```
LamboCreeper, line 35: Can you add a blank line before this return statement, please?

src/commands/InspectCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	private buildInspectEmbed(memberObj: GuildMember): MessageEmbed {
+		const embed = new MessageEmbed();
+
+		embed.setTitle(`Inspecting ${memberObj?.user.tag}`);
+		embed.setColor(memberObj?.displayColor || EMBED_COLOURS.DEFAULT);
+		embed.setThumbnail(memberObj?.user.displayAvatarURL());
+		embed.addField("User ID", memberObj?.user.id);
+		embed.addField("Username", memberObj?.user.tag);
+		if (memberObj?.nickname !== null) embed.addField("Nickname", memberObj?.nickname);
```
LamboCreeper, line 47: Can you add a blank line before this, please?

src/utils/DiscordUtil.ts
```diff
@@ -0,0 +1,31 @@
...
+	static async getGuildMember(value: string, guild: Guild): Promise<GuildMember | undefined> {
+		if (value === "") return;
+
+		// UserID
+		if ((/^[0-9]+$/g).test(value)) {
+			try {
+				return await guild.members.fetch(value);
+			} catch (error) {
```
LamboCreeper: You can remove `(error)` if you're not using it.

LamboCreeper (changes requested):
Looking really good, just a few small things.

src/utils/DiscordUtil.ts
BlackBearFTW: `Parameter name expected`

JonaVDM: You should also remove ()

BlackBearFTW:
Thank you for the review everyone, this should be all done now. But if you see anything, let me know!

LamboCreeper (approved):
Awesome! Your best work yet, great job Bear! 🙂

Jason2605 (approved):
👍

PR #145, opened 2021-01-08
Add Development Script

### LINKED ISSUE #111: Add a development script
To run the bot in development we first have to transpile to JavaScript. It would be nice if we had a `dev` script that used [ts-node](https://www.npmjs.com/package/ts-node) to transpile and run on the fly, such as `"dev": "ts-node src/app.ts"`.

JonaVDM:
Implements #111 

Adds in a script in the package.json that runs the bot via ts-node and auto reloads with nodemon

3 dependencies have been added for this to work:
- ts-node
- nodemon
- cross-env

### DIFF
--- a/src/factories/CommandFactory.ts
+++ b/src/factories/CommandFactory.ts
@@ -1,12 +1,13 @@
 import { commands_directory } from "../config.json";
 import Command from "../abstracts/Command";
 import DirectoryUtils from "../utils/DirectoryUtils";
+import { DEVELOPMENT_ENV } from "../config.json";
 
 class CommandFactory {
 	private commands: any = {};
 
 	async loadCommands(): Promise<void> {
-		const commandFileEnding = process.env.MODE === "development" ? "Command.ts" : "Command.js";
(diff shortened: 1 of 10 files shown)


src/factories/CommandFactory.ts
```diff
@@ -1,14 +1,17 @@
 import { commands_directory } from "../config.json";
 import Command from "../abstracts/Command";
 import DirectoryUtils from "../utils/DirectoryUtils";
+import { DEVELOPMENT_ENV } from "../config.json";
 
 class CommandFactory {
 	private commands: any = {};
 
 	async loadCommands(): Promise<void> {
+		const commandFileEnding = process.env.NODE_ENV === DEVELOPMENT_ENV ? "Command.ts" : "Command.js";
```
ajsaraujo: Note that this is the same logic from `app.ts:24`. Why not extract this to an util? Something like...
```ts
class DirectoryUtils {
    appendFileExtension(fileName: string): string {
        const extension = process.env.NODE_ENV === DEVELOPMENT_ENV ? ".ts" : ".js";
        return fileName + extension;
    }
}

LamboCreeper (approved):

PR #146, opened 2021-01-08
Bulk deletion

### LINKED ISSUE #131: Log Messages Which Are Bulked Deleted
**As a** moderator
**I want to** be able to see messages which are deleted in bulk
**So that I** can see what the Anti Spam Bot deleted

--- 

**Event Trigger:** `messageDeleteBulk`

For each message returned in the collection, an embed should be sent to the logs channel. This logic should be very similar the `messageDelete` logic.

Jason2605:
# Bulk deletion
Resolves #131 
## Summary
This PR adds a new handler for the bulk deletion event sent by discord. The logic for this handler is exactly the same as it is for the single message deletion, so shared logic was pulled out into a new base abstract class.

### Testing
To test this i first setup a dummy command that triggers the bulk deletion, I also entered a log channel id relevant to my test guild in the configuration file.

```ts
import {Message, TextChannel} from "discord.js";
import Command from "../abstracts/Command";

class TestCommand extends Command {
	constructor() {
		super(
			"del",
			"",
			{
				selfDestructing: false,
				aliases: ["del"]
			}
		);
	}

	async run(message: Message, args: string[]): Promise<void> {
		const channel = message.channel as TextChannel;
		// Or get messages from a specific channel
		// const channel = message.guild?.channels.resolve("<id>") as TextChannel;

		await channel.bulkDelete(5);
	}
}

export default TestCommand;
```

#### Results
Screenshot 2021-01-08 at 18 46 25

#### Note
This PR also cleans up some previous tests that were modifying `BaseMock` properties

### DIFF
--- a/src/abstracts/LogMessageDeleteHandler.ts
+++ b/src/abstracts/LogMessageDeleteHandler.ts
@@ -1,13 +1,13 @@
-import { Constants, MessageEmbed, Message, TextChannel } from "discord.js";
-import { LOG_CHANNEL_ID, EMBED_COLOURS } from "../../config.json";
-import EventHandler from "../../abstracts/EventHandler";
+import { MessageEmbed, Message, TextChannel } from "discord.js";
+import { LOG_CHANNEL_ID, EMBED_COLOURS } from "../config.json";
+import EventHandler from "./EventHandler";
 
-class LogMessageDeleteHandler extends EventHandler {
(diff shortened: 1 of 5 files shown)


LamboCreeper (approved):

LamboCreeper:
Beautiful 😍

PR #147, opened 2021-01-10
DiscordUtils bugfix

BlackBearFTW:
DiscordUtil getGuildMember only fetched first GuildMember with matching username, which led to a GuildMember "Not existing" if there were two users with the same username. 

For example:

`?inspect Jacob#7638`

Besides that I also added a test to handle nicknames.

This pull request extends #140


src/utils/DiscordUtils.ts
```diff
@@ -13,16 +13,16 @@ class DiscordUtils {
 			}
 		}
 
-		// Username + discriminator
+		// Username and/or discriminator
 		if ((/^.*#[0-9]{4}$/g).test(value)) {
 			const [username, discriminator] = value.split("#");
-			const memberList = await guild.members.fetch({query: username});
+			const memberList = await guild.members.fetch({query: username, limit: 1000000000});
```
LamboCreeper: I feel like there must be a performance loss queuing based on a limit this large. Is this just a number picked randomly? I think we could reduce it significantly.

BlackBearFTW: yep this is random, what would be a better number?

LamboCreeper (approved):

PR #150, opened 2021-01-15
GhostPingHandler now adds relevant info for deleted replies.

### LINKED ISSUE #133: Ghostping reply doesn't show original message author
**Problem:** Currently the bot does detect when a reply was deleted that pinged the original message author, but the embed doesn't show that pinged user. Which would be useful information.

reply

**Solution:** Add an if statement to check if the message was a reply, then add an extra field to the embed, that would look like:
**Replied To:**
@mention

**Optional:** Also show the content of the message that was replied to.

kaelon141:
The GhostPingHandler will now check if a deleted message is a reply to another message, and if so, add info on who was replied to and which message was replied to. This adds useful info that provides additional context for deleted replies. and directly addresses issue #133.

A notification of a deleted reply now looks like this:
image

A unit test has also been included to verify that reply info is added to the embed correctly.
image

### DIFF
--- a/src/event/handlers/GhostPingHandler.ts
+++ b/src/event/handlers/GhostPingHandler.ts
@@ -32,15 +32,16 @@ class GhostPingHandler extends EventHandler {
 				embed.setTitle("Ghost Ping Detected!");
 
 				if (repliedToUser !== null && repliedToUser !== undefined) {
-					embed.addField("Author - Reply to", `<@${message.author.id}> - <@${repliedToUser.id}>`);
+					embed.addField("Author", message.author, true);
+					embed.addField("Reply to", repliedToUser, true);
 				} else {
 					embed.addField("Author", message.author);
 				}
 
 				embed.addField("Message", message.content);
 
-				if (repliedToMessage !== null && repliedToMessage !== undefined) {
-					embed.addField("Message replied to", `https://discord.com/channels/${repliedToMessage.guild?.id}/${repliedToMessage.channel.id}/${repliedToMessage.id}`);
+				if (repliedToMessage !== null && repliedToMessage !== undefined && repliedToMessage !== null) {
+					embed.addField("Message replied to", `https://discord.com/channels/${repliedToMessage.guild?.id}/${repliedToMessage.channel.id}/${repliedToMessage.id}`, true);
 				}
 
 				embed.setFooter(`Message sent at ${DateUtils.formatAsText(message.createdAt)}`);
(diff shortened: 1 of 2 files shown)


src/event/handlers/GhostPingHandler.ts
```diff
@@ -17,9 +17,27 @@ class GhostPingHandler extends EventHandler {
 
 				const embed = new MessageEmbed();
 
+				let repliedToMessage : Message|null|undefined = null;
+				let repliedToUser: User|null|undefined = null;
```
LamboCreeper: Could you add a space either side of the pipes, just to make it easier to read, please?

src/event/handlers/GhostPingHandler.ts
```diff
@@ -17,9 +17,27 @@ class GhostPingHandler extends EventHandler {
...
+
 				embed.setTitle("Ghost Ping Detected!");
 				embed.addField("Author", message.author);
 				embed.addField("Message", message.content);
+				if (repliedToUser !== null && repliedToUser !== undefined) {
+					embed.addField("Reply to", repliedToUser);
+				}
+				if (repliedToMessage !== null && repliedToMessage !== undefined && repliedToMessage !== null) {
+					embed.addField("Message replied to", `https://discord.com/channels/${repliedToMessage.guild?.id}/${repliedToMessage.channel.id}/${repliedToMessage.id}`);
+				}
```
LamboCreeper: Could you add blank lines between each of the if statements please, i.e:
```ts
embed.addField();

if () {
}

if () {
}

embed.setFooter();
```

LamboCreeper (changes requested):
Awesome - thanks for this! Just a few small things I'd like tweaked.

Could you change it so that author and reply to are inline with each other, just so the embed takes up less space?

src/event/handlers/GhostPingHandler.ts
kaelon141: Done.

src/event/handlers/GhostPingHandler.ts
kaelon141: Done.

kaelon141:
> Awesome - thanks for this! Just a few small things I'd like tweaked.
> 
> Could you change it so that author and reply to are inline with each other, just so the embed takes up less space?

I've now altered the ghost ping embed as requested. It now looks like this:
image

The accompanying unit test has also been updated accordingly.

kaelon141:
I've now made 'Reply to' its own field again and made 'Author' and 'Reply to' inline if the deleted message is a reply. 
For reference, this what a 'normal' ghost ping notification looks like, which is unchanged:
image
And this is what a ghost ping notification looks like for a reply:
image

LamboCreeper (approved):

LamboCreeper:
Thank you so much!

PR #151, opened 2021-01-21
LogMemberLeave update/not send on raid

JonaVDM:
Changed the message in the leave embed to be name#discriminator (id) instead of a mention.

Added raid detection on the leave event, now the bot will not send a message when the user is kicked because of a raid

### DIFF
--- a/src/event/handlers/LogMemberLeaveHandler.ts
+++ b/src/event/handlers/LogMemberLeaveHandler.ts
@@ -1,6 +1,6 @@
-import { Constants, GuildMember, MessageEmbed, TextChannel } from "discord.js";
+import { Constants, GuildAuditLogs, GuildMember, MessageEmbed, TextChannel } from "discord.js";
 import EventHandler from "../../abstracts/EventHandler";
-import { EMBED_COLOURS, MEMBER_ROLE, LOG_CHANNEL_ID } from "../../config.json";
+import { EMBED_COLOURS, MEMBER_ROLE, LOG_CHANNEL_ID, RAID_SETTINGS } from "../../config.json";
 import DateUtils from "../../utils/DateUtils";
 
 class LogMemberLeaveHandler extends EventHandler {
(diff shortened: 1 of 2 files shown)


BlackBearFTW:
Sadly using your way of using the audit logs to check if a user left or was kicked, won't always work properly.

When I was also thinking of a solution, I asked in the discord.js server how I could detect if a user got kicked or just simply left.
One of the mods responded with:
 

> the only way to do so would be by fetching audit logs and hoping they arrive in time (there is no guarantee for an audit log to already be present when the GUILD_MEMBER_REMOVE event emits)

So I asked what a reasonable time was for a audit log to show up:
> there is no guarantee,
discord works based on eventual consistency - absolutely any timeout can result in wrong behaviour, granted it becomes less likely (and less usable) the higher the timeout is.

So I would handle this differently, I would pass the `joinQueue` array from `RaidDetectionHandler` to `MemberLeaveHandler`. Then when a member leaves, check if he is in the `joinQueue`. If he is, then don't send an embed.

Jason2605:
So I gave this a quick test locally. Granted what Bear is saying is true - [see here](https://discordjs.guide/popular-topics/audit-logs.html) - but in a perfect scenario it does work as expected. An issue which did however come up was a potential for a false positive. If a user was kicked as part of the raid, joined back and then left again there **could** be a scenario where this may not get logged properly.

I'm not currently sure what to suggest as an alternative to querying the audit logs and hoping they are there since discord doesn't seem to provide enough information on the given events.

BlackBearFTW:
> I'm not currently sure what to suggest as an alternative to querying the audit logs and hoping they are there since discord doesn't seem to provide enough information on the given events.

I suggested that you would pass the `joinQuery` array from RaidDetectionHandler to LogMemberLeaveHandler. Then add a check to LogMemberLeaveHandler to check if the user who left, was in that `joinQuery` array, then don't log it if he was in it.

Wouldn't that work?

Jason2605:
> > I'm not currently sure what to suggest as an alternative to querying the audit logs and hoping they are there since discord doesn't seem to provide enough information on the given events.
> 
> I suggested that you would pass the `joinQuery` array from RaidDetectionHandler to LogMemberLeaveHandler. Then add a check to LogMemberLeaveHandler to check if the user who left, was in that `joinQuery` array, then don't log it if he was in it.
> 
> Wouldn't that work?

Events are currently setup in a way that they are independent of all other events - this means to "pass" the array from one handler to another we'd either need a major overhaul in how the events work, or some sort of global state. I think before attempting either of those routes it's probably worth seeing how effective the audit log method is and make a conscious decision with that knowledge.

LamboCreeper:
Closing due to inactivity.

PR #152, opened 2021-02-24
Added ?article command, which gets the 5 latest articles from the CodeSupport Website

BlackBearFTW:
This is a preview:

This feature adds 3 files:
- ArticleCommand.ts
- ArticlePreview.ts (Interface)
- ArticleService.ts
- All the necessary tests

### DIFF
--- a/src/interfaces/CodeSupportArticle.ts
+++ b/src/interfaces/CodeSupportArticle.ts
@@ -1,16 +1,7 @@
 import CodeSupportUser from "./CodeSupportUser";
+import CodeSupportArticleRevision from "./CodeSupportArticleRevision";
 
-export default interface CodeSupportArticleRevision {
-	articleId: number;
-	content: string;
-	createdBy: CodeSupportUser;
-	createdOn: number;
-	description: string;
-	id: number;
-	// eslint-disable-next-line
-}
-
-export default interface CodeSupportArticle {
+interface CodeSupportArticle {
 	createdBy: CodeSupportUser;
 	createdOn: number;
 	id: number;
@@ -19,5 +10,6 @@ export default interface CodeSupportArticle {
 	titleId: string;
 	updatedBy: CodeSupportUser;
 	updatedOn: number;
-	// eslint-disable-next-line
 }
+
+export default CodeSupportArticle;
--- a/src/interfaces/CodeSupportArticle.ts
+++ b/src/interfaces/CodeSupportArticle.ts
@@ -1,6 +1,6 @@
-import { CodeSupportUser } from "./CodeSupportUser";
+import CodeSupportUser from "./CodeSupportUser";
 
-export interface CodeSupportArticleRevision {
+export default interface CodeSupportArticleRevision {
 	articleId: number;
 	content: string;
 	createdBy: CodeSupportUser;
@@ -9,7 +9,7 @@ export interface CodeSupportArticleRevision {
 	id: number;
 }
 
-export interface CodeSupportArticle {
+export default interface CodeSupportArticle {
 	createdBy: CodeSupportUser;
 	createdOn: number;
 	id: number;
--- a/src/interfaces/CodeSupportArticle.ts
+++ b/src/interfaces/CodeSupportArticle.ts
@@ -1,5 +1,5 @@
 import CodeSupportUser from "./CodeSupportUser";
-/*eslint-disable */
+
 export default interface CodeSupportArticleRevision {
 	articleId: number;
 	content: string;
@@ -7,6 +7,7 @@ export default interface CodeSupportArticleRevision {
 	createdOn: number;
 	description: string;
 	id: number;
+	// eslint-disable-next-line
 }
 
 export default interface CodeSupportArticle {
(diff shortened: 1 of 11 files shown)


src/services/ArticleService.ts
```diff
@@ -0,0 +1,46 @@
...
+		const { data } = await axios.get(url);
+
+		if (data.length !== 0) {
+			let articleData = data;
+
+			articleData.response.sort((a: any, b: any) => b.createdOn - a.createdOn);
+
+			articleData = articleData.response.slice(0, amount);
+
+			const articles = articleData.map((article: any) => ({
```
LamboCreeper: Can we not use a custom schema, please? We designed the schema so we should follow it, not like it's an external service.

src/services/ArticleService.ts
```diff
@@ -0,0 +1,46 @@
...
+
+	static getInstance(): ArticleService {
+		if (!this.instance) {
+			this.instance = new ArticleService();
+		}
+
+		return this.instance;
+	}
+
+	async getLatest(amount: number): Promise<ArticlePreview[]> {
```
LamboCreeper: Could this be `getLatestArticles` please?

src/services/ArticleService.ts
```diff
@@ -0,0 +1,46 @@
...
+		const url = "https://api.codesupport.dev/article/v1/articles?publishedonly=true";
+
+		const { data } = await axios.get(url);
+
+		if (data.length !== 0) {
+			let articleData = data;
+
+			articleData.response.sort((a: any, b: any) => b.createdOn - a.createdOn);
+
+			articleData = articleData.response.slice(0, amount);
```
LamboCreeper: I believe you could do this all as one:
```js
const articleData.response
    .sort()
    .slice()
```

Also, no use of `any` please :)

LamboCreeper (changes requested):

src/services/ArticleService.ts
BlackBearFTW: Okay, what should I replace any with? `Object` ?

LamboCreeper: Type it :) You can see the schemas: https://api.codesupport.dev/swagger-ui.html

src/services/ArticleService.ts
BlackBearFTW: Mhm, GithubService also doesn't do `getGithubIssues`. It does `getIssues`.... why should this have the name of the service in the method?

LamboCreeper: Fair enough, you can keep it as is then.

src/interfaces/CodeSupportArticle.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface CodeSupportArticleRevision {
+	articleId: number;
+	content: string;
+	createdBy: CodeSupportUser;
+	createdOn: number;
+	description: string;
+	id: number;
+}
+
+export interface CodeSupportArticle {
```
LamboCreeper: Can this be the default export, please?

src/interfaces/CodeSupportUser.ts
```diff
@@ -0,0 +1,16 @@
+export interface CodeSupportRole {
+	code: string;
+	id: number;
+	label: string;
+}
+
+export interface CodeSupportUser {
```
LamboCreeper: Can this be the default export, please?

src/services/ArticleService.ts
```diff
@@ -0,0 +1,41 @@
...
+		return this.instance;
+	}
+
+	public async getLatest(amount: number): Promise<CodeSupportArticle[]> {
+		const url = "https://api.codesupport.dev/article/v1/articles?publishedonly=true";
+
+		const { data } = await axios.get(url);
+
+		if (data.length !== 0) {
+			data.response.sort((a: CodeSupportArticle, b: CodeSupportArticle) => b.createdOn - a.createdOn);
```
LamboCreeper: I don't think this is doing anything - you don't assign it to a variable and it doesn't pass by reference.

LamboCreeper (changes requested):
Nice, just some small things!

src/services/ArticleService.ts
```diff
@@ -0,0 +1,42 @@
...
+		if (data.length !== 0) {
+			return data.response
+				.sort((a: CodeSupportArticle, b: CodeSupportArticle) => b.createdOn - a.createdOn)
+				.slice(0, amount);
+		}
+
+		return [];
+	}
+
+	public buildArticleURL(title: string): string {
```
LamboCreeper, line 33: This could be static.

src/services/ArticleService.ts
```diff
@@ -0,0 +1,42 @@
...
+		return `https://codesupport.dev/article/${title}`;
+	}
+
+	public buildProfileURL(user: string): string {
```
LamboCreeper: This could be static.

LamboCreeper: Actually, this method has nothing directly to do with articles. It should really live on a `WebsiteUserService`.

LamboCreeper (changes requested):

LamboCreeper (approved):

PR #155, opened 2021-03-02
DiscordMessageLinkHandler

Jason2605:
# DiscordMessageLinkHandler
## Summary
This PR changes `DiscordMessageLinkHandler` to cater for an issue brought up where it was not correctly handling new lines, it also fixes an issue where the previous handler only ever worked in the context of a single message link within a given message.

##### Single message with other text
Screenshot 2021-03-02 at 19 52 15

##### Multiple links, still caters for escaped links too
Screenshot 2021-03-02 at 19 52 24

##### Multiple links
Screenshot 2021-03-02 at 19 52 34

## Note
This PR had to change the tsconfig to allow for `matchAll` along with an issue regarding iterating.

### DIFF
--- a/src/event/handlers/DiscordMessageLinkHandler.ts
+++ b/src/event/handlers/DiscordMessageLinkHandler.ts
@@ -15,7 +15,7 @@ class DiscordMessageLinkHandler extends EventHandler {
 			const index = match.index;
 
 			if (index !== undefined) {
-				let link = match[0];
+				let [link] = match;
 
 				if (message.content.charAt(index - 1) !== "<" || message.content.charAt(index + link.length) !== ">") {
 					link = link.replace(/app/, "").replace(/ptb\./, "");
--- a/src/event/handlers/DiscordMessageLinkHandler.ts
+++ b/src/event/handlers/DiscordMessageLinkHandler.ts
@@ -8,21 +8,22 @@ class DiscordMessageLinkHandler extends EventHandler {
 	}
 
 	async handle(message: Message): Promise<void> {
-		const messageRegex = /https:\/\/(ptb\.)?discord(app)?\.com\/channels\//gm;
-		const linkIndex = message.content.search(messageRegex);
+		const messageRegex = /https:\/\/(?:ptb\.)?discord(?:app)?\.com\/channels\/\d+\/\d+\/\d+/gm;
+		const matches = [...message.content.matchAll(messageRegex)];
 
-		if (linkIndex === -1) return;
+		for (const match of matches) {
+			const index = match.index;
 
-		const wordIndex = message.content.slice(0, linkIndex).split(" ").length - 1;
-		const linkContent = message.content.split(" ")[wordIndex];
+			if (index !== undefined) {
+				let link = match[0];
 
-		if (linkContent.startsWith("<") && linkContent.endsWith(">")) return;
-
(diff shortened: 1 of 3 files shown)


src/event/handlers/DiscordMessageLinkHandler.ts
```diff
@@ -8,21 +8,23 @@ class DiscordMessageLinkHandler extends EventHandler {
...
+		const matches = message.content.matchAll(messageRegex);
 
-		if (linkIndex === -1) return;
+		for (const match of matches) {
+			const index = match.index;
 
-		const wordIndex = message.content.slice(0, linkIndex).split(" ").length - 1;
-		const linkContent = message.content.split(" ")[wordIndex];
+			if (index !== undefined) {
+				let link = match[0];
```
LamboCreeper: You can do `let [link] = match;` here

src/event/handlers/DiscordMessageLinkHandler.ts
```diff
@@ -8,21 +8,23 @@ class DiscordMessageLinkHandler extends EventHandler {
...
 
-		if (linkContent.startsWith("<") && linkContent.endsWith(">")) return;
+				if (message.content.charAt(index - 1) !== "<" || message.content.charAt(index + link.length) !== ">") {
+					link = link.replace(/app/, "").replace(/ptb\./, "");
+					const messagePreviewService = MessagePreviewService.getInstance();
```
LamboCreeper, line 22: I think it makes sense to declare variables at the top of the block when possible.

test/event/handlers/DiscordMessageLinkHandlerTest.ts
```diff
@@ -25,9 +25,22 @@ describe("DiscordMessageLinkHandler", () => {
 			handler = new DiscordMessageLinkHandler();
 		});
 
+		it("sends a message in message channel when contains discord message link mid sentence", async () => {
+			const message = CustomMocks.getMessage();
+			const channel = CustomMocks.getTextChannel();
```
LamboCreeper: Since you're declearing these in each `it` you could just do it in the `before`.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
LamboCreeper, line 3: What's the difference between ES5 and ES2015?

src/event/handlers/DiscordMessageLinkHandler.ts
Jason2605: This point i disagree with. You generally want to declare variables in the lowest scope possible, it makes reasoning about your code a lot simpler.
- It's closer to it's usage so it's easier to understand
- It will leave scope when it's actually meant to rather than lingering around for longer than it should (includes added GC bonus since it can be collected quicker, but in our case that probably doesn't really matter)

tsconfig.json
Jason2605: ES2015 is ES6, ES5 is ES2009.

src/event/handlers/DiscordMessageLinkHandler.ts
LamboCreeper: I personally think it makes the code look messier, but if it has performance benefits I'll allow it.

LamboCreeper (approved):

PR #158, opened 2021-03-27
Using .get for IDs instead of .find

Murky-MoMo:
<Collection>.find() (similar to <Array>.find(<callback>) )iterates through the collection to get the channel, whereas <Collection>.get("ID") is <Map>.get(<Key>)
it's more eficient


BlackBearFTW:
You need to fix the test since its still stubbing the find method:
https://github.com/codesupport/discord-bot/blob/master/test/event/handlers/LogMessageSingleDeleteHandlerTest.ts#L31

Same at line 42

PR #160, opened 2021-03-29
InviteCycler concept

BlackBearFTW:
So I made another prevention against those raiders. This handler basically uses the maxUse invite feature to create a new invite when the old one expires. This would change a public invite out when its expired, meaning that you have to use a new invite link, this will stop the raiders from joining in masses.

This needs some setup:
- Disallow **anyone** from making invites.
- Delete all existing invites (the bot should be off if you do this)

Then you can startup the bot,
- Make an invite to right channel yourself (most likely #rules-and-info in our case)
- Delete that invite

Now the bot wil create an invite itself and set the invite designation to that channel. it will also send a message in that channel containing the invite url (that message will also be edited with the new invite when it gets created)

This still a concept and probably needs some work, even if this concept gets accepted, I still need to write tests.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -94,5 +94,10 @@
             "description": "Keep it appropriate, some people use this at school or at work."
         }
     ],
-    "MAX_INVITE_USAGE": 1
+    "INVITE": {
+        "MAX_USAGE": 1,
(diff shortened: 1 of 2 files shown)


LamboCreeper:
Closing due to inactivity.

PR #161, opened 2021-03-30
Made raid warning message look more professional by putting it in an embed

BlackBearFTW:
I made the raid warning message look more professional by putting it inside an embed, so it stands out more in the general chat.
opinions

Besides that I also added the time in the footer so you can see when the raid took place.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -32,7 +32,8 @@
     "EMBED_COLOURS": {
         "SUCCESS": "#35BC31",
         "ERROR": "#BC3131",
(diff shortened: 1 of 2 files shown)


BlackBearFTW:
Currently it looks different, since the red color specifies it being an error, so a warning color was added, now it looks like this:
embed

LamboCreeper:
> Currently it looks different, since the red color specifies it being an error, so a warning color was added, now it looks like this:

I don't believe you have pushed up this change?

LamboCreeper (approved):

PR #164, opened 2021-04-05
Feature: command ?projects

Bas-Sarrechia:
enables the ?projects command

inside the assets folder there is a projects.json file. This serves as pseudo database (currently populated with 15 different project ideas). Each project is tagged with different searchable attributes under the common name 'tags'.

Functional tests:
?project <- yields instructions
?project qsfkljqsdlkfjsdlfj <- yields an error to refine the search result
?project beginner <- yields a beginner project with a green boarder
?project medium spa <- yields a medium project for spa development with an orange border
?project spa medium <- same result set as above

### DIFF
--- a/src/commands/ProjectCommand.ts
+++ b/src/commands/ProjectCommand.ts
@@ -2,7 +2,7 @@ import Command from "../abstracts/Command";
 import {Message, MessageEmbed} from "discord.js";
 // @ts-ignore
 import projects from "../../assets/projects.json";
-import {Project} from "../interfaces/Project";
+import Project from "../interfaces/Project";
 
 export default class ProjectCommand extends Command {
 	private readonly defaultSearchTags = ["easy", "medium", "hard"];
--- a/src/commands/ProjectCommand.ts
+++ b/src/commands/ProjectCommand.ts
@@ -4,10 +4,11 @@ import {readFile} from "fs";
 
 export default class ProjectCommand extends Command {
 	private readonly fileDirectory = "./assets/projects.json";
+
 	constructor() {
 		super(
 			"project",
-			"Returns a random project idea based on given parameters.",
+			"Returns a random project idea based on given parameters."
 		);
 	}
 
@@ -33,7 +34,7 @@ export default class ProjectCommand extends Command {
 
 					embed.setColor(this.loadDifficultyColorMap().get(difficulty || "") || "#add8e6");
 					embed.setTitle(project.title);
-					embed.addFields({name: "tags", value: project.tags.join(" ")},);
+					embed.addFields({name: "tags", value: project.tags.join(" ")});
 					embed.addField("Project details", project.description);
 				} else {
 					embed.setTitle("Could not find a project");
@@ -45,16 +46,13 @@ export default class ProjectCommand extends Command {
 		}
 	}
 
-    private readonly loadDifficultyColorMap: () => Map<string, string> = () => new Map([["easy", "#35BC31"], ["medium", "#ffa500"], ["hard", "#bc3131"]]);
-    private readonly removeTooLongDescriptions: (project: Project) => boolean = ({description}) => description.length <= 2048;
-    private readonly filterTags: (project: Project, requestedTags: Array<string>) => boolean =
-        ({tags}, requestedTags: Array<string>) => requestedTags.every(tag => tags.includes(tag))
-    private readonly retrieveFirstFoundTag: (project: Project, tagsToRetrieve: Array<string>) => string | undefined =
-        ({tags}, tagsToRetrieve: Array<string>) => tagsToRetrieve.filter(tag => tags.map(tag => tag.toLowerCase()).includes(tag)).pop();
-}
+	private readonly loadDifficultyColorMap: () => Map<string, string> = () => new Map([["easy", "#35BC31"], ["medium", "#ffa500"], ["hard", "#bc3131"]]);
 
-interface Project {
-    title: string,
-    tags: Array<string>,
-    description: string,
+	private readonly removeTooLongDescriptions: (project: Project) => boolean = ({description}) => description.length <= 2048;
+
+	private readonly filterTags: (project: Project, requestedTags: Array<string>) => boolean =
+		({tags}, requestedTags: Array<string>) => requestedTags.every(tag => tags.includes(tag));
+
+	private readonly retrieveFirstFoundTag: (project: Project, tagsToRetrieve: Array<string>) => string | undefined =
+		({tags}, tagsToRetrieve: Array<string>) => tagsToRetrieve.filter(tag => tags.map(tag => tag.toLowerCase()).includes(tag)).pop();
 }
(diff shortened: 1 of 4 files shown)


src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
+import Command from "../abstracts/Command";
+import {Message, MessageEmbed} from "discord.js";
+import {readFile} from "fs";
```
LamboCreeper: Why read the file rather than just import it? You could just do:
```ts
import projects from "../../assets/projects.json";
```

Doing so would also reduce a lot of logic you have within the class.

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	async run(message: Message, args: string[]): Promise<void> {
+		const embed = new MessageEmbed();
+		const query = args.map(arg => arg.toLowerCase()).filter(arg => arg.trim().length > 0);
+
+		if (args.length === 0) {
+			embed.setTitle("Projects")
+				.setColor("#add8e6")
+				.setDescription("Please provide arguments on the projects command.");
```
LamboCreeper: Project convention isn't to chain embed creation methods, can this be changed please.

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+					.filter(project => this.filterTags(project, query))
+					.sort(() => 0.5 - Math.random()).pop();
+
+				if (project) {
+					const difficulty = this.retrieveFirstFoundTag(project, ["easy", "medium", "hard"]);
+
+					embed.setColor(this.loadDifficultyColorMap().get(difficulty || "") || "#add8e6")
+						.setTitle(project.title)
+						.addFields({name: "tags", value: project.tags.join(" ")},)
+						.addField("Project details", project.description);
```
LamboCreeper: Same as above.

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+				} else {
+					embed.setTitle("Could not find a project")
+						.setColor("#bc3131")
+						.setDescription("try to enter less search arguments to broaden your search.");
```
LamboCreeper: Same as above.

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+				.setColor("#add8e6")
+				.setDescription("Please provide arguments on the projects command.");
+			await message.channel.send(embed);
+		} else {
+			readFile(`${this.fileDirectory}`, {encoding: "utf8"}, async (error, projectData) => {
+				const projects: Array<Project> = JSON.parse(projectData);
+				const project = projects
+					.filter(this.removeTooLongDescriptions)
+					.filter(project => this.filterTags(project, query))
+					.sort(() => 0.5 - Math.random()).pop();
```
LamboCreeper: I'm curious, what was your thought process being randomising the results?

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+
+				if (project) {
+					const difficulty = this.retrieveFirstFoundTag(project, ["easy", "medium", "hard"]);
```
LamboCreeper: Thoughts on making a `ProjectDifficulty` enum rather than magic strings?

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+			});
+		}
+	}
+
+    private readonly loadDifficultyColorMap: () => Map<string, string> = () => new Map([["easy", "#35BC31"], ["medium", "#ffa500"], ["hard", "#bc3131"]]);
+    private readonly removeTooLongDescriptions: (project: Project) => boolean = ({description}) => description.length <= 2048;
+    private readonly filterTags: (project: Project, requestedTags: Array<string>) => boolean =
+        ({tags}, requestedTags: Array<string>) => requestedTags.every(tag => tags.includes(tag))
+    private readonly retrieveFirstFoundTag: (project: Project, tagsToRetrieve: Array<string>) => string | undefined =
+        ({tags}, tagsToRetrieve: Array<string>) => tagsToRetrieve.filter(tag => tags.map(tag => tag.toLowerCase()).includes(tag)).pop();
```
LamboCreeper: Could this logic be made neater please, it's quite hard to read all shoved together like this?

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+interface Project {
```
LamboCreeper: Can this live inside the interfaces directory, please?

LamboCreeper (changes requested):
Nice, thanks for the contribution. Just a couple of questions around decisions made and some stylistic changes.

src/commands/ProjectCommand.ts
Bas-Sarrechia: the resultset can be expansive, if a user would merely search on 'beginner' or 'easy', there could be 10 different projects applying. If there is no randomisation, they'll always retrieve the same project. By randomising, a lot more options are to be shown, and a lot more 'projects' can share the same tags.

src/commands/ProjectCommand.ts
Bas-Sarrechia: I had considered an enum. However, I decided against it ultimately and decided that a difficulty is ultimately just another tag of the project. This reduced complexity by just having a tag. The easy / medium/ hard thing will just provide a neat border / search option.

LamboCreeper: Hmm okay, that makes sense I suppose. I'm just not sure I like the magic strings - could they at least be constants?

src/commands/ProjectCommand.ts
Bas-Sarrechia: Agreed (and good point, I didn't know imports like this worked in javascripty stuff). On contrary, are you sure you want this perma loaded in memory?

LamboCreeper: Yeah, should be fine loaded.

Bas-Sarrechia: Resolved by exposing a (rather ugly) function which provides the project data, which in turn is stubbed by the tests. if you know a better way to stub the import lmk

src/commands/ProjectCommand.ts
```diff
@@ -0,0 +1,60 @@
+import Command from "../abstracts/Command";
+import {Message, MessageEmbed} from "discord.js";
+// @ts-ignore
```
LamboCreeper, line 3: Would you mind adding to this comment explaining why the `@ts-ignore` is needed?

src/interfaces/Project.ts
```diff
@@ -0,0 +1,5 @@
+export interface Project {
```
LamboCreeper: Might as well export this as the default.

LamboCreeper (changes requested):

src/commands/ProjectCommand.ts
Bas-Sarrechia: File '......' is not under 'rootDir' '...........'. 'rootDir' is expected to contain all source files.

LamboCreeper (approved):

LamboCreeper (approved):

PR #168, opened 2021-04-10
Added new filetypes to the whitelist filter, since discord now embeds most files. 

BlackBearFTW:
Discord now embeds most files in auto generated code blocks.
txtfile

So I after some research these are filetypes that tend to get uploaded alot in the server, but these would get embedded if accepted by the filter.

List of new added files:
``` "webm",
        "txt",
        "py",
        "html",
        "css",
        "js",
        "jsx",
        "ts",
        "tsx",
        "jar",
        "java",
        "php",
        "c",
        "cpp",
        "cxx",
        "cs",
        "lua",
        "dart",
        "json",
        "sql",
        "md",
        "sh",
        "bat",
        "log"```
        
        however here are all languages that get embedded from now on
        https://gist.github.com/BlackBearFTW/26a957cafa0e0238f8b3aea087127194


LamboCreeper (approved):
Nice one, thanks Bear. :)

PR #169, opened 2021-04-11
Project command updated and ProjectService Added

BlackBearFTW:
Hey, 

Made some changes to the Project Command and added a ProjectService

First commit:

`?project`  without any query now returns any random project.
`?project #react`, tags with # infront now also work as query.

Second Commit

Extracted all Project code to a ProjectService so the project command only has to take care of making it look nice and so it fits better with the rest of the codebase.

ProjectService has 3 methods:
- formatTags
- getProjectByTags
- getDifficulty 

Todo: 
- Figure out how to mock the projects json file so I can write more tests for the ProjectService


LamboCreeper:
Closing due to inactivity.

PR #171, opened 2021-04-15
Revert "Added new filetypes to the whitelist filter, since discord now embeds most files. "

theboxmage:
Reverts codesupport/discord-bot#168

Per user requests on the server, discord embeds don't seem to be any better than requesting a codeblock. Can this be removed?


LamboCreeper (approved):
Approving reversion since it seems to be that most people do not like this change. Open to add it back in the future if we see it viable.

PR #173, opened 2021-05-22
Added a minesweeper command.

BlackBearFTW:
Hi,

I added a minesweeper command, which generates an embed with a random 11x11 grid.

Knipsel

It also has 3 difficulties:
- Easy
- Medium
- Hard

`?minesweeper [easy|medium|hard]`

This minesweeper was implemented as service.
Lastly also added a NumberUtils to pick a random number in a certain range.

### DIFF
--- a/src/commands/MineSweeperCommand.ts
+++ b/src/commands/MineSweeperCommand.ts
@@ -15,24 +15,21 @@ class MineSweeperCommand extends Command {
 	async run(message: Message, args: string[]) {
 		const embed = new MessageEmbed();
 		const mineSweeperService = MineSweeperService.getInstance();
+		const givenDifficulty = args[0].toLowerCase();
 
 		if (!args[0]) args[0] = "easy";
 
-		interface DifficultyLevel {
-			[key: string]: {ratio: number}
-		}
-
-		const difficultyLevels: DifficultyLevel = {
-			"easy": {ratio: 7},
-			"medium": {ratio: 5},
-			"hard": {ratio: 3}
+		const difficultyLevels: {[key: string]: number} = {
+			"easy": 7,
+			"medium": 5,
+			"hard": 3
 		};
 
-		if (Object.keys(difficultyLevels).includes(args[0].toLowerCase())) {
-			const difficulty = difficultyLevels[args[0].toLowerCase()];
+		if (Object.keys(difficultyLevels).includes(givenDifficulty)) {
+			const difficulty = difficultyLevels[givenDifficulty];
 
-			embed.setTitle(`MineSweeper (${StringUtils.capitalise(args[0].toLowerCase())})`);
-			embed.setDescription(mineSweeperService.generateGame(11, 11, difficulty.ratio));
+			embed.setTitle(`MineSweeper (${StringUtils.capitalise(givenDifficulty)})`);
+			embed.setDescription(mineSweeperService.generateGame(difficulty));
 			embed.setColor(EMBED_COLOURS.DEFAULT);
 		} else {
 			embed.setTitle("Error");
--- a/src/commands/MineSweeperCommand.ts
+++ b/src/commands/MineSweeperCommand.ts
@@ -15,9 +15,30 @@ class MineSweeperCommand extends Command {
 		const embed = new MessageEmbed();
 		const mineSweeperService = MineSweeperService.getInstance();
 
-		embed.setTitle("MineSweeper");
-		embed.setDescription(mineSweeperService.generateGame(9, 9, 5));
-		embed.setColor(EMBED_COLOURS.DEFAULT);
+		if (!args[0]) args[0] = "easy";
+
+		interface DifficultyLevel {
+			[key: string]: {display: string, ratio: number}
+		}
+
+		const difficultyLevels: DifficultyLevel = {
+			"easy": {display: "Easy", ratio: 7},
+			"medium": {display: "Medium", ratio: 5},
+			"hard": {display: "Hard", ratio: 3}
+		};
+
+		if (Object.keys(difficultyLevels).includes(args[0].toLowerCase())) {
+			const difficulty = difficultyLevels[args[0].toLowerCase()];
+
+			embed.setTitle(`MineSweeper (${difficulty.display})`);
+			embed.setDescription(mineSweeperService.generateGame(11, 11, difficulty.ratio));
+			embed.setColor(EMBED_COLOURS.DEFAULT);
+		} else {
+			embed.setTitle("Error");
+			embed.setDescription("You must provide an existing difficulty");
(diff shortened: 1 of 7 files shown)


src/commands/MineSweeperCommand.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		interface DifficultyLevel {
+			[key: string]: {ratio: number}
+		}
+
+		const difficultyLevels: DifficultyLevel = {
+			"easy": {ratio: 7},
+			"medium": {ratio: 5},
+			"hard": {ratio: 3}
+		};
```
LamboCreeper: Why is this an object of objects? The inner objects only ever have one key, you could just simplify this to:
```suggestion
		const difficultyLevels: DifficultyLevel = {
			easy: 7,
			medium: 5,
			hard: 3
		};
```

src/commands/MineSweeperCommand.ts
```diff
@@ -0,0 +1,48 @@
...
+
+	async run(message: Message, args: string[]) {
+		const embed = new MessageEmbed();
+		const mineSweeperService = MineSweeperService.getInstance();
+
+		if (!args[0]) args[0] = "easy";
+
+		interface DifficultyLevel {
+			[key: string]: {ratio: number}
+		}
```
LamboCreeper: Don't define an interface within a method or class. Also, refer to the comment below about changing the signature.

src/commands/MineSweeperCommand.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		const difficultyLevels: DifficultyLevel = {
+			"easy": {ratio: 7},
+			"medium": {ratio: 5},
+			"hard": {ratio: 3}
+		};
+
+		if (Object.keys(difficultyLevels).includes(args[0].toLowerCase())) {
```
LamboCreeper: You're doing `args[0].toLowerCase()` multiple times through out this, why not set that to a variable and use that throughout instead?

src/commands/MineSweeperCommand.ts
```diff
@@ -0,0 +1,48 @@
...
+			const difficulty = difficultyLevels[args[0].toLowerCase()];
+
+			embed.setTitle(`MineSweeper (${StringUtils.capitalise(args[0].toLowerCase())})`);
+			embed.setDescription(mineSweeperService.generateGame(11, 11, difficulty.ratio));
```
LamboCreeper: Magic Numbers: What is `11`, `11`? I would make them static class properties of `GRID_WIDTH` and `GRID_HEIGHT`.

src/services/MineSweeperService.ts
```diff
@@ -0,0 +1,77 @@
...
+	public generateGame(rows: number, cols: number, ratio: number): string {
+		const bombCount = rows * cols / ratio;
+		const bombIndex = [];
+		const numbers = [":zero:", ":one:", ":two:", ":three:", ":four:", ":five:", ":six:", ":seven:", ":eight:"];
+		const positions = [[0, -1], [0, 1], [-1, -1], [-1, 0], [-1, 1], [1, -1], [1, 0], [1, 1]];
+		const bomb = ":boom:";
+
+		const grid = Array.from(Array(rows), () => new Array(cols).fill(0));
+
+		function isOutsideEdges(rowIndex: number, colIndex: number) {
```
LamboCreeper: Don't declare a function within a method - make this a private method on the class.

src/services/MineSweeperService.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		return this.instance;
+	}
+
+	public generateGame(rows: number, cols: number, ratio: number): string {
+		const bombCount = rows * cols / ratio;
+		const bombIndex = [];
+		const numbers = [":zero:", ":one:", ":two:", ":three:", ":four:", ":five:", ":six:", ":seven:", ":eight:"];
+		const positions = [[0, -1], [0, 1], [-1, -1], [-1, 0], [-1, 1], [1, -1], [1, 0], [1, 1]];
```
LamboCreeper: I don't like this variable name, what are the positions of - can you be more specific?

LamboCreeper (changes requested):
Nice, thanks for adding a bit of fun. Just a few questions/things I would change.

src/services/MineSweeperService.ts
```diff
@@ -0,0 +1,79 @@
...
+
+				grid[rowI + rowPosition][colI + colPosition] += 1;
+			});
+		});
+
+		for (let rowI = 0; rowI < this.GRID_ROWS; rowI++) {
+			for (let colI = 0; colI < this.GRID_COLUMNS; colI++) {
+				// eslint-disable-next-line no-continue
+				if (typeof grid[rowI][colI] !== "number") continue;
+				grid[rowI][colI] = numbers[grid[rowI][colI]];
```
JonaVDM: Since this `if` only effects one line, and disables eslint in doing it. Maybe refactor the if to not use continue?

Maybe something like 
```ts
if (typeof grid[rowI][colI] === "number") {
    grid[rowI][colI] = numbers[grid[rowI][colI]];
}
```

src/services/MineSweeperService.ts
```diff
@@ -2,6 +2,8 @@ import NumberUtils from "../utils/NumberUtils";
 
 class MineSweeperService {
 	private static instance: MineSweeperService;
+	private readonly GRID_ROWS = 11;
+	private readonly GRID_COLUMNS = 11;
```
LamboCreeper: These can be static.

src/services/MineSweeperService.ts
```diff
@@ -0,0 +1,87 @@
...
+
+			grid[randomRow][randomColumn] = MineSweeperService.BOMB;
+			bombPositions.push([randomRow, randomColumn]);
+			i++;
+		}
+
+		return [grid, bombPositions];
+	}
+
+	private calculateBombSurrounding(grid: any[][], bombPositions: any[][]) {
```
LamboCreeper: Don't use `any`.

src/services/MineSweeperService.ts
```diff
@@ -0,0 +1,87 @@
...
+				if (typeof grid[rowI + rowPosition][colI + colPosition] !== "number") return;
+
+				grid[rowI + rowPosition][colI + colPosition] += 1;
+			});
+		});
+
+		return grid;
+	}
+
+	private formatGameResult(grid: any[][]) {
```
LamboCreeper: Don't use `any`.

LamboCreeper (changes requested):

LamboCreeper (approved):
Thanks Bear! 💣💥

PR #175, opened 2021-05-25
Support whitelisting of channel commands.

LamboCreeper:
#### Overview
- Adds support to whitelist a command's execution to given channels
- Whitelisted `?minesweeper` to #bot-chat


test/event/handlers/CommandParserHandlerTest.ts
```diff
@@ -107,6 +107,17 @@ describe("CommandParserHandler", () => {
 			expect(errorWasThrown).to.be.false;
 		});
 
+		it("should not run command if whitelisted to a channel which is not the execution channel", async () => {
+			const message = BaseMocks.getMessage();
+			const command = new MockWhitelistedCommand();
+			const commandSpy = sandbox.spy(command, "run");
+
+			message.channel.id = "abcdef";
```
Jason2605, line 115: We want to avoid modifying properties of `BaseMock` as they're static - realise this is probably taken from the above test but we should change that one too

LamboCreeper:
No longer required, migrating to the Interactions API will allow us to natively handle this.

PR #177, opened 2021-06-02
Added duck command

BlackBearFTW:
image

For when people need a little help searching on the internet


LamboCreeper:
We’ve discussed this feature as a Moderation team and decided we do not want it a part of our bot. Whilst we may occasionally joke to people about using services like LMGTFY we do not want to encourage people getting help from third parties, that defeats our purpose.

PR #191, opened 2021-06-24
bugfix/RaidDetectionHandler

### LINKED ISSUE #190: bug/RaidDetection slow down warning messages
### Overview
Please describe the bug below:
Currently the bot spams general full with warnings, I have a solution for this problem, by adding in a queue_counter that only sends the message every x amount of dealt with queues

### Expected Behaviour
Please describe how you expect this functionality to work:
It should slow down and not send that message as often

### Actual Behaviour
Please describe how this functionality is actually working:
As I said, general now gets spammed multiple times back to back.

BlackBearFTW:
Resolves #190 

### Overview
- Refactored RaidDetectionHandler.
- Added cycle_counter functionality, so that warning message gets send less often.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -26,7 +26,7 @@
     },
     "RAID_SETTINGS": {
         "TIME_TILL_REMOVAL": 60,
-        "MAX_QUEUE_SIZE": 10,
+        "MAX_QUEUE_SIZE": 10
     },
     "EMBED_COLOURS": {
         "SUCCESS": "#35BC31",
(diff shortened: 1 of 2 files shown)


BlackBearFTW:
Okay so I changed the logic, instead of the try/catch block executing for each member (as long as there are more then 10 members in queue). It now only runs the the try/catch block once, when the joinQueue is bigger then the minimal raid size, it will then kick everyone in the queue, no matter how big the queue is.

It will also divide the queue into groups, for example with a queue size of 30 and a raid being every 10 members (this is the RAID_SETTINGS.MAX_QUEUE_SIZE), this will send 1 message for every 10 members (so 3 messages in total)

LamboCreeper:
Closing due to inactivity.

PR #197, opened 2021-07-29
Fix ghost pinging bots being detected

### LINKED ISSUE #196: Necessity of detecting ghost pinging bots
Currently, ghost pinging by deleting a message mentioning a bot (only bot or author + bot) is detected.
Deleting a reply (ping = on) to a bot is detected as well.

Is that the intended behavior? If not, I'd like to fix it.

ChrisCh7:
Bots don't care if they're ghost pinged, so it shouldn't be detected.

Resolves #196 

### Overview
- Make the ghost ping not be detected when author only mentions himself and/or bots, and when author replies to a bot.
- Add test cases which check the scenarios when author pings himself and bot, pings just bots, replies to bot.

### DIFF
--- a/src/event/handlers/GhostPingHandler.ts
+++ b/src/event/handlers/GhostPingHandler.ts
@@ -13,7 +13,7 @@ class GhostPingHandler extends EventHandler {
 			if (!message.author?.bot) {
 				const usersMentioned = message.mentions.users;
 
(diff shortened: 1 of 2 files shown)


LamboCreeper (approved):
Excellent! Thank you so much!

PR #200, opened 2021-08-06
Impersonation Handler

LamboCreeper:
### Overview
Bans people from changing their names to the names of the moderators.


Bas-Sarrechia (commented):
you can make use of .localeCompare(string, undefined, { sensitivity: 'accent'}) === 0; for string case insensitive comparison. Alternative a levenshtein distance calculation can be done in order to mend things like 'Iambo' the L is an i. 

Additional req - banwordlist -> iffy ʎɟɟᴉ itsiffy
additionally this should take into account the join date. people who joined x time before should not be affected

saramaebee:
also maybe check for people joining with the same name as a mod?

theboxmage:
Is there work to be done on this? I haven't seen much communication on this feature in a while.  If something is left, I'd be willing to try.

PR #195, opened 2021-07-16
Feature/v13 base update

BlackBearFTW:
Resolves #194 

**NOTE: Some things might still be broken**

### Overview
Please bullet point the changes you have made below:
- Changed the way embeds were send
- Change field types to string
- Changed intends.ALL to intends (stored in config file) (Intends.ALL was removed, because of lazy developers)

PR #201, opened 2021-08-08
Discord.js upgrade to version 13.0.1!

### LINKED ISSUE #194: Support V13
_Please describe the functionality below:_
This PR will fix the changes described on [here](https://deploy-preview-680--discordjs-guide.netlify.app/additional-info/changes-in-v13.html)

this will not include the slash commands, buttons or select menu's

_Please explain why this functionality would benefit our community below:_
This will keep the bot up to date with the changes made to the discord api and discord.js library

JonaVDM:
Resolves #194 

### Overview
Based on Bears PR #195
- Made the app actually start (needs Node version >16)
- Fixed test


src/event/handlers/LogMessageBulkDeleteHandler.ts
```diff
@@ -7,10 +7,10 @@ class LogMessageBulkDeleteHandler extends LogMessageDeleteHandler {
 	}
 
 	async handle(messages: Collection<Snowflake, Message>): Promise<void> {
-		for (const message of messages.array()) {
+		messages.forEach(async message => {
 			await super.sendLog(message);
-		}
+		});
```
LamboCreeper: You could simplify this even more to: `await Promise.all(messages.map(super.sendLog))`.

src/utils/DiscordUtils.ts
```diff
@@ -26,6 +26,11 @@ class DiscordUtils {
 
 		return fetchedMembers.first();
 	}
+
+	static getAllIntents(): BitFieldResolvable<IntentsString, number> {
+		// Stole... copied from an older version of discord.js...
```
LamboCreeper, line 31: It would be great if you could link to the actual file you got this from.

test/commands/AdventofcodeCommandTest.ts
```diff
@@ -76,7 +76,7 @@ describe("Adventofcode Command", () => {
 			sandbox.stub(command, "getYear").returns(2019);
 
 			await command.run(message, ["2021"]);
-			const embed = messageMock.getCall(0).firstArg.embed;
+			const embed = messageMock.getCall(0).firstArg.embeds[0];
```
LamboCreeper, line 79: If you wanted you could change all of these to:
```suggestion
			const [embed] = messageMock.getCall(0).firstArg.embeds;
```

Not particularly bothered though.

test/utils/DiscordUtilsTest.ts
```diff
@@ -3,6 +3,7 @@ import { createSandbox, SinonSandbox } from "sinon";
 import DiscordUtils from "../../src/utils/DiscordUtils";
 import { BaseMocks, CustomMocks } from "@lambocreeper/mock-discord.js";
 import { Collection, GuildMemberManager } from "discord.js";
+import assert from "assert";
```
LamboCreeper: Don't believe this is being used anywhere?

src/abstracts/Command.ts
```diff
@@ -12,7 +12,7 @@ abstract class Command {
 		this.options = options;
 	}
 
-	abstract async run(message: Message, args?: string[]): Promise<void>;
+	abstract run(message: Message, args?: string[]): Promise<void>;
```
LamboCreeper, line 15: What was the reasoning behind removing `async` here?

src/abstracts/EventHandler.ts
```diff
@@ -5,7 +5,7 @@ abstract class EventHandler {
 		this.event = event;
 	}
 
-	abstract async handle(...args: any[]): Promise<void>;
+	abstract handle(...args: any[]): Promise<void>;
```
LamboCreeper, line 8: What was the reasoning behind removing `async` here?

LamboCreeper (commented):
Thanks for this Bear / Jona / Jason - really appreciate it. Just a few questions and minor suggestions.

One thing which does bother me is the amount of casting, though I'm happy to address that in a different pull request.

BlackBearFTW:
> One thing which does bother me is the amount of casting, though I'm happy to address that in a different pull request.

This is indeed an issue, sadly discord.js now has specific types for everything and doesn't allow strings anymore, so somewhere somehow the type needs to change from string to the required type.

src/abstracts/Command.ts
BlackBearFTW: I did this, typescript doesn't allow `async` to be used with `abstract`

> TS1243: 'async' modifier cannot be used with 'abstract' modifier.

src/abstracts/EventHandler.ts
BlackBearFTW: I did this, typescript doesn't allow async to be used with abstract

TS1243: 'async' modifier cannot be used with 'abstract' modifier.

test/commands/AdventofcodeCommandTest.ts
BlackBearFTW: would that not give you an array of embeds? instead of just one, like we have it now?

src/abstracts/EventHandler.ts
JonaVDM: In an abstract `async` does not need to be specified, the `promise<void>` is already enough

test/commands/AdventofcodeCommandTest.ts
JonaVDM: those are a lot of things that need to be changed, rather not. Maybe later

LamboCreeper (approved):

PR #207, opened 2021-11-03
Slash Command Support Added

BlackBearFTW:
No I wont follow your template, I was bored and you needed slash commands.

I will not write tests and I will not adapt all the commands, if you want me to do that, you can first donate at my paypal.

Now leave me alone, you aren't even brave enough to dm me yourself Lambo, you had TBM do it.

PS: Matthias I did read the documentation, fuck you in particular for telling me that I didn't do something.


BlackBearFTW:
For people wondering "Why doesn't it build", well its because every single command needs to be totally changed to allow interactions instead of a message object in the run method, which then also needs to call interaction.reply()

I don't currently have the time to do that, this PR is just the groundwork on updating the commands.

BlackBearFTW:
I will now close this PR, because ofcourse having a single interaction in the past 3 months with one of you, TBM.

Gets me called out as being a dogshit developer and that I don't deserve any respect.

Go fuck yourself y'all , you haven't fucked changed (Except the people that I actually enjoyed talking to, you know who you are, stay cool)

- Bear

kayomn:
What on Earth am I reading?

PR #210, opened 2022-01-07
Add Inherited Config to Support Multi-Environment Configuration

LamboCreeper:
A common development pain for this project has been the fact you have to edit and then undo your changes to `config.json` between development and pushing to GitHub. To combat this, I have introduced [@codesupport/inherited-config](https://github.com/codesupport/inherited-config) to allow contributors to add a `config.dev.json` file to overwrite any config options which relate just to their environment.


Jason2605 (approved):
Nice 👍

PR #214, opened 2022-01-31
Hiring or looking

theboxmage:
Resolves Existing Non-Slash commands

### Overview
- Converted ?hl to a slash command

Don't approve this without telling me if you have a better command than /hiring

### DIFF
--- a/src/commands/slash/HiringLookingCommand.ts
+++ b/src/commands/slash/HiringLookingCommand.ts
@@ -1,11 +1,11 @@
 import {ColorResolvable, CommandInteraction, MessageEmbed} from "discord.js";
+import {Discord, Slash} from "discordx";
 import getConfigValue from "../../utils/getConfigValue";
 import GenericObject from "../../interfaces/GenericObject";
-import {Discord, Slash} from "discordx";
 
 @Discord()
 class HiringLookingCommand {
-	@Slash("hiring")
+	@Slash("hl")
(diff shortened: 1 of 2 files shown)


src/commands/slash/HiringLookingCommand.ts
```diff
@@ -1,17 +1,12 @@
...
-		super(
-			"hl",
-			"Get information on how to correctly format a Hiring/Looking post"
-		);
-	}
-
-	async run(message: Message): Promise<void> {
+@Discord()
+class HiringLookingCommand {
+	@Slash("hiring")
```
LamboCreeper: I would keep this as `hl` instead of `hiring` for the sake of muscle memory.

src/commands/slash/HiringLookingCommand.ts
```diff
@@ -1,17 +1,12 @@
-import {ColorResolvable, Message, MessageEmbed} from "discord.js";
-import Command from "../../abstracts/Command";
+import {ColorResolvable, CommandInteraction, MessageEmbed} from "discord.js";
 import getConfigValue from "../../utils/getConfigValue";
 import GenericObject from "../../interfaces/GenericObject";
+import {Discord, Slash} from "discordx";
```
LamboCreeper: Imports from packages should be above imports from local files.

LamboCreeper (approved):

PR #217, opened 2022-01-31
Conversion of Inspect to slash command

theboxmage:
### Overview
- Converts ?inspect to /Inspect
- Adds tests for slash command
- Adds a single change to app.ts to include the new slash command.

### DIFF
--- a/src/commands/slash/InspectCommand.ts
+++ b/src/commands/slash/InspectCommand.ts
@@ -8,15 +8,10 @@ import DiscordUtils from "../../utils/DiscordUtils";
 @Discord()
 class InspectCommand {
 	@Slash("inspect")
-	async onInteract(@SlashOption("id", {description: "ID of user", required: false}) userID: string, interaction: CommandInteraction): Promise<void> {
-		let memberID: string;
-
-		if (userID === undefined) {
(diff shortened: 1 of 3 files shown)


src/commands/slash/InspectCommand.ts
```diff
@@ -1,28 +1,23 @@
...
-	async run(message: Message, args: string[]) {
-		const userObj = args.length > 0
-			? await DiscordUtils.getGuildMember(args[0], message.guild!)
-			: message.member!;
+@Discord()
+class InspectCommand {
+	@Slash("inspect")
+	async onInteract(
+		@SlashOption("user", {type: "MENTIONABLE", required: false}) userID: GuildMember,
+			interaction: CommandInteraction): Promise<void> {
```
LamboCreeper, line 13: I think this would look neater if it was formatted as:
```suggestion
	async onInteract(
		@SlashOption("user", {type: "USER", required: false}) userID: GuildMember,
		interaction: CommandInteraction
	): Promise<void> {
```

Also, changing `MENTIONABLE` to `USER` so that it doesn't autocomplete roles too.

LamboCreeper (approved):

PR #222, opened 2022-02-07
Rule command

theboxmage:
Converts existing command

### Overview
- Updates rule command and rule command's tests


src/commands/slash/RuleCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Discord()
+class RuleCommand {
+	@Slash("rule")
+	async onInteract(
+		@SlashOption("rule", {
+			autocomplete: (interaction: AutocompleteInteraction) => {
+				const rule = getConfigValue<Rule[]>("rules").map(it => ({name: it.name, value: it.triggers[0]}));
+
+				interaction.respond(rule);
```
LamboCreeper: I don't really like how this callback lives within the decorator. I wonder if it would be better to read the rules in separately and store them in a `const`?

LamboCreeper (approved):

PR #252, opened 2022-06-22
Feature/242 ghost ping update handler

### LINKED ISSUE #242: Ghost Ping not detected on message edit
### Overview
Please describe the bug below:
When you edit a message that contains a ping and remove the mention, it doesn't trigger ghost ping.
### Expected Behaviour
Please descripe how you expect this functionaltiy to work:
This should show the ghost ping embed.
### Actual Behaviour
Please describe how this functionality is actually working:
The user still gets the ping, but since the message was edited, it doesn't show where they were pinged.

olivier-b-be:
Resolves #242 

### Overview
- Added the GhostPingUpdateHandler to detect removed mentions on an edited message
- Renamed the GhostPingHandler to GhostPingDeleteHandler as it specifically deals with the deleted messages
  - Updated `setFooter` on the embed to remove the deprecation warning


src/event/handlers/GhostPingUpdateHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+		const removedMentions = oldMentionedUsers.filter(user => {
+			for (const [, newUser] of newMessage.mentions.users) {
+				if (user.id === newUser.id) {
+					return false;
+				}
+			}
+
+			return true;
+		});
```
LamboCreeper: I don't really like this if I'm honest. I think something like the below would be neater:
```suggestion
		const removedMentions = oldMentionedUsers.filter(user => !newMessage.mentions.users.includes(user.id));
```

src/event/handlers/GhostPingUpdateHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+		if (removedMentions.size === 0 || removedMentions.every(user => user.id === oldMessage.author.id || user.bot)) return;
+
+		const button = new MessageButton();
+
+		button.setLabel("View Edited Message");
+		button.setStyle("LINK");
```
LamboCreeper: Is there a constant available in Discord.js that we can use here?

LamboCreeper (commented):
Great stuff Banana, I really appreciate it. Just two minor things.

src/event/handlers/GhostPingUpdateHandler.ts
olivier-b-be: There is indeed! Did not think to check, will change it to that.

src/event/handlers/GhostPingUpdateHandler.ts
olivier-b-be: Done!

LamboCreeper (approved):

PR #254, opened 2022-06-26
Convert legacy Advent Of Code command to slash command 

### LINKED ISSUE #244: Bot crashes on AOC command
### Overview
When there is a invalid cookie-token for the advent of code api and someone requests the stats of a user, the bot crashes instead of showing an error.

image
image

### Expected Behaviour
Please descripe how you expect this functionaltiy to work:
An error should be presented to the user.

### Actual Behaviour
Please descripbe how this fucntionality is actually working:
The bot just crashes.

olivier-b-be:
Converts the last legacy command to slash command and removes legacy commands functionality

Resolves #244 

### Overview
- Converted the legacy command to a slash command with two optional parameters
  - `year`: If a valid year is passed in it will use that year to either get the leaderboard of that year or the user statistics from that year
  - `name`: If a name is passed in that is found in the leaderboard ranking of a year it will return the user's statistics
- Slightly updated the visuals so that an anonymous user's id is also padded correctly in the leaderboard that is shown and added the year values for displaying info:

| Before | After |
| ----------- | ----------- |
| image      | image|

- Removal of the legacy command handling way
  - Removed the `legacy` folders from `src/commands` and `test/commands` as the last command is now converted to a slash command
  - Deleted the CommandFactory and CommandParserHandler including the tests as they are no longer used for any (legacy) commands).
  - Moved the files in the `commands/slash` directory to the `commands` directory and updated the imports
- Updated README commands section to showcase the new way that commands are created for this bot


README.md
```diff
@@ -52,23 +52,27 @@ If you would like to overwrite values in `config.json` to better suit your local
...
 
 ### Creating Commands
-To create a command, create a new file in `src/commands` named `<CommandName>Command.ts`. Commands should extend the `Command` abstract and `super` the command name, the description and any command options (if applicable). When a command is run, it triggers the `run` method. This method has two parameters:
-- `message` - The Discord Message object
-- `args` - An optional array of strings that the user sends along with the command
+To create a command, create a new file in `src/commands` named `<CommandName>Command.ts`. DiscordX is used to register the commands as slash commands using decorators. Commands should have the `@Discord()` decorator above the class name. The command should have an `onInteract` async function that is decorated using `@Slash`. In `@Slash` decorator's parameters you have to pass in a name which will be the name of the command when used in Discord, it has an optional options parameter where you can for instance pass in a description.
```
LamboCreeper, line 55: Thanks for updating this!

src/commands/AdventOfCodeCommand.ts
```diff
@@ -0,0 +1,134 @@
...
+
+		if (!!year && year <= yearToQuery) {
+			yearToQuery = year;
+		} else if (!!year) {
+			await interaction.reply({embeds: [this.errorEmbed(`Year requested not available.\nPlease query a year between 2015 and ${yearToQuery}`)], ephemeral: true});
+			return;
+		}
+
+		const link = `https://adventofcode.com/${yearToQuery}/leaderboard/private/view/${getConfigValue<string>("ADVENT_OF_CODE_LEADERBOARD")}`;
+		const description = `Leaderboard ID (${currentAOCYear}): \`${getConfigValue<string>("ADVENT_OF_CODE_INVITE")}\`\n\n[View Leaderboard (${yearToQuery})](${link})`;
```
LamboCreeper: We could use a Button component for the link now.

LamboCreeper (commented):
Woohoo! Migration complete! Just a comment around switching to a button now.

src/commands/AdventOfCodeCommand.ts
olivier-b-be: Added! The description is now simply the leaderboard ID, with the button linking to the leaderboard.

olivier-b-be:
Incase you want a final check on the end product.

image

LamboCreeper (approved):
Thanks, Banana, great stuff!

PR #257, opened 2022-07-30
Added a helpful usage field for users that try to use the project command

### LINKED ISSUE #256: Include helpful description in /project error message
When a user uses the `/project` command it's possible the person has no idea what to use as a `query` for the parameter. It'd be really helpful to list all the difficulties and some random tags for the user to use the next time. Right now it simply says it couldn't find a result for the given query which is meaningless if you don't know what terms to use.

People trying to find a project to work on would get instant feedback on how to work with the `/project` command, wouldn't have to wait on other people in the server to either tell them what to look for or use the command first themselves.

olivier-b-be:
Resolves #256 

### Overview
- Updated the usage field when the embed returns an error

image


LamboCreeper (approved):
Looks good, thanks!

PR #292, opened 2024-05-19
Regular Role Logging

LamboCreeper:
### Overview
Statbot adds and removes a `Regular` role based on users activity. I want to be alerted of this, so that I can easily monitor. This introduces a change which sends a message to a channel when a user is given or loses this role.


saramaebee (approved):
lgtm, good change to have

PR #297, opened 2024-08-06
Anonymous Feedback Form Command

LamboCreeper:
Adding a new `/feedback` command which allows users to submit anonymous feedback.


src/commands/FeedbackCommand.ts
theboxmage, line 1: Imports on lines 3-8 don't match the style we use in other files. They're usually on the same line.

src/commands/FeedbackCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+		const embed = new EmbedBuilder()
+
+		embed.setTitle("New Anonymous Feedback");
+		embed.setDescription(feedbackInput);
+
+		const message = await feedbackChannel?.send({
+			embeds: [embed]
+		});
+
+		await message.startThread({
```
theboxmage, line 50: Who has access to this thread? Mods? Or do the users have access as well.

LamboCreeper: Just mods, yep.

theboxmage (approved):

PR #314, opened 2026-01-10
Added Spam account ban counter, and Pinging via /rule

bogdan-glitchm:
### Overview
Changes:
- There is now a counter for the number of spam accounts banned
- You can now pass an argument to the `/rule` command to also ping a certain person (cooldown of 30s per pinged person and of 3m per pinger)


cornusandu:
btw tests and linting passed. I’ll wait for a review

cornusandu:
As of the previous commit, coverage has now been raised from 94.8% (before the PR) to 95.11%

PR #329, opened 2026-06-12
chore: update dependencies

### LINKED ISSUE #328: Update dependencies
Update NodeJS version + dependencies

Keep the bot dependencies up-to-date, requested by Lambo

ChrisCh7:
Resolves #328

### Overview
- Update NodeJS version + dependencies

> [!IMPORTANT]
> Used Claude to do the changes, please ensure there are no problems before merging.

<details><summary>Changes description from Claude convo</summary>

The upgrade is complete and fully verified: lint, build, all 197 tests, and coverage pass on Node 26, with 0 npm audit vulnerabilities.

What changed

Runtime (Node 18 → 26): .nvmrc, engines, and the Dockerfile now target Node 26. Both GitHub workflows previously didn't pin Node at all (they silently used the runner default) — they now use actions/setup-node@v6 with node-version-file: ".nvmrc" and npm caching; I also bumped actions/checkout v1 → v6 while in there.

Dependencies: everything to latest — TypeScript 6.0.3, discord.js 14.26.4, discordx 11.13.3, dotenv 17, tsyringe 4.10, reflect-metadata 0.2, ESLint 10.5, typescript-eslint 8.61 (meta-package replaces the two old @typescript-eslint/* packages), mocha 11.7.6 + ts-mocha 11.1, sinon 22, nyc 18, cross-env 10. Two deliberate holds: chai stays at 4.5.0 (chai 5+ is ESM-only and would break this CommonJS project) and ts-node stays at 10.9 (required by ts-mocha). The lockfile was regenerated (now lockfileVersion 3).

ESLint flat config: .eslintrc.js is replaced by eslint.config.js, which keeps eslint-config-codesupport via FlatCompat as agreed. The lint script drops the now-removed --ext flag, and two eslint-disable comments that ESLint 10 flags as unused were removed.

Code fallout (all small):
- tsconfig.json: module/moduleResolution moved from commonjs/Node to node16 — TS 6 errors on the deprecated node10 resolution; output is still CommonJS.
- Four message.channel.send(...) call sites got an isSendable() guard — discord.js 14.16+ added PartialGroupDMChannel (which has no .send) to the channel union.
- NewUserAuthenticationHandlerTest fixtures: reaction emojis now use id: null (correct for unicode emojis) — newer discord.js resolves custom-emoji ids through client.application, which is null on a mock client.

Two things worth flagging:
1. patch-package was added with a one-line patch to @lambocreeper/mock-discord.js 3.0.0 (its latest version): it constructs channels without passing the client, relying on a guild.client fallback that newer discord.js removed. The patch applies automatically via postinstall (verified with a clean npm ci), and the Dockerfile copies patches/ before npm ci. The proper long-term fix is a new mock-discord.js release — upstream CodeSupport still pins discord.js 14.8 and hasn't hit this yet.
2. npm overrides force mocha's transitive diff → 8.x and serialize-javascript → 7.0.5, clearing all three audit findings (mocha still pins vulnerable ranges). I verified failing-test diff rendering still works with the override.

A cosmetic note: dotenv 17 prints an "injected env" banner on startup/tests — silence it with DOTENV_CONFIG_QUIET=true if it bothers you.

</details>


patches/@lambocreeper+mock-discord.js+3.0.0.patch
LamboCreeper, line 1: We should probably look at removing this as a dependency as I don't really plan on updating the library.

LamboCreeper (approved):
Thank you!

